PR #106, opened 2013-07-10
(Premature) optimization of nqp::join() on JVM

gerdr:
Instead of using a `StringBuilder` (which may need to repeatedly re-allocate its internal buffer), we allocate a character array of correct size. This improves the performance of micro-benchmarks, but the effect on setting compilation is minimal.

### DIFF
--- a/src/vm/jvm/runtime/org/perl6/nqp/runtime/Ops.java
+++ b/src/vm/jvm/runtime/org/perl6/nqp/runtime/Ops.java
@@ -2676,25 +2676,44 @@ public final class Ops {
     }
     
     public static String join(String delimiter, SixModelObject arr, ThreadContext tc) {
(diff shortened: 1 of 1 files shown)


moritz:
I've done some rudimentary benchmarks in https://gist.github.com/moritz/ab4e7cd01c36032e4d6b

Results:
* startup time unchanged (2.02 +- 0.06 => 2.08 +- 0.15)
* performance improvements of 9.6%, but in my numbers that's only three standard deviations (so don't trust it too much).

So a decent improvement in a micro benchmark while only slightly increasing code size and complexity sounds like an overall win to me.

rurban:
See also https://github.com/parrot/parrot/issues/1123 to fix this in the parrot backend.

PR #215, opened 2015-01-03
Add support for loading bare .class files and increase visibility for a few BootJavaInterop fields and inner class constructors.

peschwa:
The increased visibility is a neccessity for the branch that improves JVM-Interop on Rakudo, while loading of bare .class files was just incomplete.


moritz:
Not knowing JVM much, I cannot comment on this much. But it makes no new spectest fails here, so I'd be willing to merge it if nobody objects.

PR #218, opened 2015-01-15
Do not try to intify a missing argument

usev6:
return zero instead and let the check for the right number of arguments
do its job later on

fixes RT #122907


moritz:
Thanks!

PR #219, opened 2015-01-24
Change message for mismatch between specified arguments in directives an...

usev6:
...d supplied arguments

cmp. http://irclog.perlgeek.de/perl6/2015-01-24#i_9996416


moritz:
Thanks @usev6. I've given you a commit bit for NQP.

usev6:
Thanks @moritz!

PR #225, opened 2015-03-09
Fix EOF with newlines on 32k byte boundaries

### LINKED ISSUE #224: nqp-j readline wrongly sets eof if newline is on 32k boundary
nqp-j wrongly sets EOF if a newline is on a 32k boundary. Demo:

    perl6 -e'say "a" x 32767 for 1 .. 4;' > data
    nqp-j -e'my $i := 0; my $fh := nqp::open("data", "r"); until nqp::eoffh($fh) {nqp::readlinefh($fh); $i++}; nqp::say($i);' # 1
    # or with a perl6-j:
    perl6-j -ne'$i++; END {say $i}' data # 1

 This does not happen when reading from standard input:

    cat data | perl6-j -e'my $i = 0; loop {$*IN.get; last if $*IN.eof; $i++}; say $i;' # 4

This happens at a different place with other data sets. 'seq 1 10240' causes a problem at 6775. (`head -n 6775 data | wc -l` gives 32768)

This is nqp version 2015.02-75-g3d325cf built on JVM / This is perl6 version 2015.02-234-g7759845 built on JVM
openjdk version "1.8.0_31"
OpenJDK Runtime Environment (build 1.8.0_31-b13)
OpenJDK 64-Bit Server VM (build 25.31-b07, mixed mode)
Fedora 21 x64

avuserow:
Fixes #224 by changing the EOF checking to use the size of the file
rather than attempting to read a character using getc(). This fixes a
problem which happens when a line ends on a multiple of 32768 bytes.

I tested this via NQP's make test, ran a spectest on Rakudo (which had several failures but none that looked related), and tested manually. I only tested on regular files on Linux.

### DIFF
--- a/src/vm/jvm/runtime/org/perl6/nqp/io/FileHandle.java
+++ b/src/vm/jvm/runtime/org/perl6/nqp/io/FileHandle.java
@@ -87,11 +87,8 @@ public class FileHandle extends SyncHandle implements IIOSeekable {
             return false;
         else {
             try {
-                long position = fc.position();
-                getc(tc);
-                fc.position(position);
-                readBuffer = null;
-                return false;
+                eof = fc.position() >= fc.size();
+                return eof;
             }
             catch (Exception e) {
                 eof = true;


FROGGS:
Thank you very very much!

PR #228, opened 2015-03-13
Look for <terminator> in HLL::Grammar.nulltermish

usev6:
Fixes RT #79174


usev6:
In RT #79174 it is noted that HLL::Grammar.nulltermish misses a check for <?terminator>.

I'm not sure whether my patch is the right way to add that check. It looks a bit arkward. On the other hand with this patch the following code works:

```perl6-m -e 'my @a = 1, 2, 3, given 42; say @a'
1 2 3```

usev6:
RT #79174 got fixed in a clean way with commit 67179841ac

PR #232, opened 2015-04-13
Unlink tmp file after tests

paultcochrane:
Clean up the automatically generated "tmp" file which is created while running the test suite.


moritz:
@paultcochrane thanks, I've also added you to the nqp team, so you can push directly in future

PR #248, opened 2015-11-01
Feature/repeat regex all seq

ungrim97:
Support ^10 & ^* ranges in regex quantifiers

### DIFF
--- a/src/QRegex/P6Regex/Actions.nqp
+++ b/src/QRegex/P6Regex/Actions.nqp
@@ -162,6 +162,7 @@ class QRegex::P6Regex::Actions is HLL::Actions {
         }
         else {
             my $min := $<min>.ast;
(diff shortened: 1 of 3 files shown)


peschwa:
http://irclog.perlgeek.de/perl6/2015-11-01#i_11464806 has some discussion regarding allowing ^* as regex quantifier.

ungrim97:
After some discussions it would appear that:

1: ^* should likely not be supported and ideally use of ^* and ^Inf should throw a warning informing the user they likely want the * quantifier.

2: ^10 current in this patch has an off by one error and acts like 0..10. This is incorrect and should act as 0..^10.

3: Due to the above it should be possible to support the currently unimplemented non inclusive ranges (1^..10, 1..^10 and 1^..^10)

PR #273, opened 2016-01-24
Bump MOAR_REVISION to fix windows building

johnspurr:
There was a problem that stopped moar from building on windows because it couldn't find msinttypes/inttypes.h, but that problem was fixed in this revision.


johnspurr:
I'm not entirely sure on the formatting of a MOAR_REVISION bump, I noticed that the number between the release in the commit number changes, but I'm not sure how so I left it alone. This request was mainly about bumping the commit version

niner:
The MOAR_REVISION is created by "git describe" run in the moar repo. The number you referred to is the number of commits since the base tag (2015.12).

johnspurr:
Thanks niner, I fixed it using git describe this time.

niner:
Thanks! I squashed the two commits into one and pushed the result.

PR #280, opened 2016-04-25
enable custom jars to be used for building the JVM backend

tomboy-64:
This patch enables custom paths provided as classpaths for the 3rd party components.


tomboy-64:
I just noticed, I'm not done. There are still the standard jars being installed.

tomboy-64:
Alright, works now.

PR #281, opened 2016-04-27
Enable usage of custom asm.jar, asm-tree.jar, jline.jar and jna.jar

tomboy-64:
This enables downstream to provide custom jars instead of the provided 3rd party ones.


peschwa:
Patch looks good, NQP and Rakudo still build. On those grounds I'm not opposed to merging, but I'd like a comment or two from packagers for other distributions. I think @paultcochrane does debian?

tomboy-64:
ping

jnthn:
No comments forthcoming, so I think we merge. It seems a step in the right direction, and can be tweaked in the future if needed.

PR #287, opened 2016-05-21
Make nqp::nextfiledir have the same output on the jvm as on moar

MasterDuke17:


MasterDuke17:
Rakudo PR #772 fixes the break in Rakudo this commit causes

PR #298, opened 2016-07-24
[RFC] remove Parrot support

moritz:
Activity levels (or the absence thereof) in Parrot in the last half a year to year indicates to me that we should give up on the idea of resurrecting the Rakudo Parrot backend, and as such I see no reason to keep the NQP backend.

This simplifies the build system, decreases the checkout size on disk, and gives fewer false positives when search for relevant things in the source code.

If no major objections from NQP or Rakudo core contributors arise, I plan to merge this pull request in about a week. Maybe earlier if the overall resonance is positive.

CC to @jnthn and @pmichaud just in case :)


arnsholt:
No objections on my part.

lizmat:
nor from me

pmichaud:
I have no objections to removing the parrot components from nqp's master branch.  Perhaps create a "parrot" branch or tag to make it easy to get to the parrot-specific stuff if someone decides to try to restore it.

Pm

coke:
After some initial negative thoughts about this, I'm on board; especially if we tag the last known commit with a parrot build on it.

coke:
There are still many references to parrot here. Mind if I add a few more removals, @moritz ?

moritz:
@coke I have no objections to a more thorough removal. Consider this a collaborative effort.

moritz:
@pmichaud ok, I'll create a branch before merging this PR.

jnthn:
I'm also fine with this. So far as I'm aware, NQP on Parrot has not built for a number of months, and I can't recall anybody complaining about that - which is probably indicative of the interest level. 

Leaving a branch behind is OK, though IMHO a tag would be more natural (tags are for marking significant points in development, whereas branches are more for ongoing development). If anybody fancied the leg work, putting that tag on the last release that actually built against Parrot would maybe be most fitting, but I sure ain't volunteering anyone for that task. :-) A tag on the last commit before we remove it work be fine by me.

moritz:
Thanks for the comments everybody, I've merged this now, and created a `parrot` tag, which is the last commit before the parrot removal.

PR #307, opened 2016-09-10
Properly numify non-ascii tokens in regex/grammars

MasterDuke17:
Fix [RT #127075](https://rt.perl.org/Ticket/Display.html?id=127075)

Passes `make test` in nqp and `make spectest` in rakudo


MasterDuke17:
Also, `perl6 -e '"abc 123" ~~ /$<número>=\d+/'`, which wasn't mentioned in the RT, now works (it used to give the same error as the example in the RT).

PR #308, opened 2016-09-27
Throw exception instead of calling dieInternal

usev6:
After removing this check for a SaveStackException
the special case for JVM in Rakudo's src/core/control.pm
is no longer needed.

There are no new failing tests (neither with 'make test'
for NQP nor with 'make test; make install; make spectest'
for Rakudo).

Fixes two tickets for rakudo-j:

* https://rt.perl.org/Ticket/Display.html?id=122732

* https://rt.perl.org/Ticket/Display.html?id=127967


peschwa:
Thanks!

PR #311, opened 2016-10-06
Fix handling of QAST::Want on JVM (2)

usev6:
There were two places where 'v' was used instead of $RT_VOID.
After changing those, we can test for void context in handling
of QAST::Want.

Fixes NPE with 'sub f { sub { say 1 } }; f()()'.


zoffixznet:
Thanks.

PR #314, opened 2016-10-25
Avoid NPE in add_cache for listy thing with hole

usev6:
Fixes RT #128320


usev6:
Actually, I'm not sure that this is the right way to fix the issue from RT #128320. Especially, I'm not sure that it makes sense to use an additional value MD_CACHE_NULL in argTup.

However, the conclusion in RT #128320 was, that the JVM specific code should protect from the null.

AlexDaniel:
https://github.com/Raku/old-issue-tracker/issues/5356 (RT #128320) seems to be closed as resolved. So I think we don't need this PR anymore. Let me know if that's not the case.

usev6:
For the record: Nearly two years after the old issue https://github.com/Raku/old-issue-tracker/issues/5356 has been closed, I've put some code very similiar to that from this PR in: https://github.com/Raku/nqp/commit/670400e625b4 and https://github.com/Raku/nqp/commit/051cbec2bc09.

I've totally forgotten about this PR. Thanks for cleaning up! ;)

PR #318, opened 2016-11-06
Remove serialize_sc

MasterDuke17:
It was mostly a no-op and only called in one place, so just inline the
actual functionality. A similar edit will have to be made for the JS backend.

Passes `make spectest`


jnthn:
Maybe @pmurias can comment further as he's just bootstrapped a backend recently, but I think the point of the current factoring is - as the comment this removes said - to make life easier for those who are porting to new backends.

pmurias:
I override the serialize_sc method when cross-compiling to both js + moar.

PR #319, opened 2016-11-15
Initial support for line directives

MasterDuke17:
This will allow backtraces to report the original file and line number
instead of m-CORE.setting and its line number.

This also works in EVALs, making it easier to know which one has died.

Passes `make test`

Once this has been merged, rakudo will require https://github.com/rakudo/rakudo/pull/919


zoffixznet:
Thanks a lot.

PR #322, opened 2016-11-28
Add bin/ to .gitignore

MasterDuke17:
So it does not accidentally get committed


coke:
Why was this necessary? I just did a build/install of nqp locally and this directory is nowhere to be seen.

MasterDuke17:
Looks like it's created when you do a JVM build.
https://github.com/perl6/nqp/blob/master/tools/build/Makefile-JVM.in#L101-L102

PR #325, opened 2016-12-02
[JVM] Add null check in backtrace generation

usev6:


usev6:
I just committed the code (without merge commit), so this can be closed.

PR #327, opened 2016-12-04
Add SQL as an output option for profiles

MasterDuke17:
This creates a .sql file with several CREATE TABLE statements and many
INSERT INTO statements. This has been tested with SQLite and with the
appropriate queries the results seem to match the JSON and HTML outputs.


zoffixznet:
Thanks a lot.

PR #333, opened 2016-12-27
Make $!orig a str in ParseShared and NQPMatch

MasterDuke17:
No noticeable effect, but hopefully the more specialized bindattr_s and
getattr_s will be more efficient.

Passes `make m-test`. Rakudo's `make m-spectest` has a couple errors, but they're all Unicode tests that seem related to the recent Unicode commits in MoarVM, NQP, Rakudo.

### DIFF
--- a/src/QRegex/Cursor.nqp
+++ b/src/QRegex/Cursor.nqp
@@ -1,7 +1,7 @@
 # Some things that all cursors involved in a given parse share.
 my class ParseShared is export {
     has $!CUR_CLASS;
-    has $!orig;
+    has str $!orig;
     has str $!target;
     has int $!highwater;
     has @!highexpect;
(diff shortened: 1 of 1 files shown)


moritz:
I don't know if this is relevant, but Rakudo's Match.orig can be a non-Str (for example when matching against an Int). We should make sure this continues working with this pull request.

MasterDuke17:
Ah, are there any tests for this in roast?

moritz:
I couldn't find any, so I added some in roast@ea2ac7f272a2e285343fc697521d0481f85a6088.

MasterDuke17:
not ok 39 - non-Str orig

# Failed test 'non-Str orig'
# at t/spec/S05-capture/match-object.rakudo.moar line 84
# expected: 12345
#      got: "12345"
ok 40 - .prematch on non-Str
ok 41 - .postmatch on non-Str

jnthn:
:-1: to this, for the reason @moritz mentioned. We already have `$!target` which is the `str` form of `$!orig` and used for many places that need the performance. I don't think `$!orig` is used anything like so much, so all the places that could benefit from an optimization like this already are.

moritz:
Sorry for having you go through this work with insufficient test coverage :(

MasterDuke17:
No worries, that's why I made this a PR instead of committing directly, I had my doubts. And hey, we got some useful tests written because of it.

PR #340, opened 2017-01-10
Add a method cache to NFA building

MasterDuke17:
This seems like a hack, but it reduces the time to (just) define 50 custom operators from 35s to 28s.

Passes `make m-test` and a Rakudo built with it passes `make m-spectest`.

### DIFF
--- a/src/QRegex/NFA.nqp
+++ b/src/QRegex/NFA.nqp
@@ -65,8 +65,10 @@ class QRegex::NFA {
 
     has $!known;
 
+    has %!method_cache;
+
     method new() {
-        my $new := self.bless(:states(nqp::list()), :edges(0), :LITEND(0), :known([]));
+        my $new := self.bless(:states(nqp::list()), :edges(0), :LITEND(0), :known([]), :method_cache(nqp::hash()));
         $new.addstate();  # storage for fates
         $new.addstate();  # entry point, mostly fanout epsilons
         $new;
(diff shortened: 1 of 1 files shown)


lizmat:
I'd really like to see jnthn's feedback on this.

MasterDuke17:
Calling @jnthn and @timo (who I was chatting about this with) into the fray. I'll admit I was kind of surprised it worked, and there's probably a better overall solution, but I PR'ed this to see what people thought.

jnthn:
First of all, nice find! :-)

The reason we don't just `nqp::findmethod` is because of the `:notrace(1)` that we also pass to `find_method`. That in turn is because of the `--rxtrace` feature. As I think was discovered on channel (but not noted here), you can actually replace this `find_method` call with the `nqp::findmethod` op and get a huge win. This is true. Unfortunately, it will also bust the `--rxtrace` feature. At the same time, we shouldn't slow down everything for the sake of a rarely used feature.

A better approach might be to introspect the meta-object and check if tracing is turned on or not (`$cursor.HOW.traced($cursor)` will return non-zero if so). If it isn't, use nqp::findmethod. If it is, use the slow-path `find_method`. Since the `traced` method is tiny, spesh will inline it, so there's probably not much cause to cache the result of it.

jnthn:
I think this wants closing, as we merged an alternative solution?

zoffixznet:
Yeah. It was https://github.com/perl6/nqp/commit/de709396b6c267144e8a40d510764b7a91c90662

PR #341, opened 2017-01-11
Use nqp::findmethod instead of *.HOW.find_method

MasterDuke17:
If the meta-object is not being traced we can use the much faster NQP
op. This reduces the time needed to compile a script that just defines
50 custom operators from ~60s to ~40s.

Passes `make m-test` and a Rakudo built with it passes `make m-spectest`.

jnthn++, timotimo++, lizmat++


jnthn:
Looks good to me; thanks!

PR #343, opened 2017-01-24
Support Nd's in sprintf format strings

MasterDuke17:
Passes `make m-test` and a Rakudo built with this passes `make m-spectest`.

### DIFF
--- a/src/HLL/sprintf.nqp
+++ b/src/HLL/sprintf.nqp
@@ -483,13 +483,13 @@ my module sprintf {
         }
 
         method idx($/) {
-            my $index := $<param_index> - 1;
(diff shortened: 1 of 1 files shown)


MasterDuke17:
A quick and dirty benchmark (not of sprintf in particular, but of nqp::radix). 20m iterations of nqp::radix($foo) takes ~5.6s on my machine. 20m iterations of +$foo takes ~4.6s.

samcv:
@MasterDuke17 did you do any performance benchmarking on this? IMO it would be best to be able to use radix as many places as possible and then if needed fast path ascii numbers in the radix function

lizmat:
The grammar parsing of printf will drown out any performance issues this patch brings, I think

PR #344, opened 2017-01-25
make nfa debugger resistant to negative indents

timo:
in at least one case $ind reached far into the negatives, and
before this patch it'd just straight-up crash during debug output.


timo:
noticed this while looking at https://rt.perl.org/Ticket/Display.html?id=130637 - i wasn't able to spot anything obvious going wrong there, though.

zoffixznet:
Merged per jnthn++: https://irclog.perlgeek.de/perl6-dev/2017-01-31#i_14021668

PR #348, opened 2017-03-21
Update qregex tests to allow backend specific todo's

samcv:
Previously :pge<comment> would todo under all backends. This has been changed
to :any<comment>. It will also allow todoing under specific backends as well
(jvm and moar currently supported).


samcv:
PR to make sure this does not break any JVM tests in unforseen ways using Travis CI.

samcv:
Travis CI is passing on JVM and MoarVM now. Merging.

PR #349, opened 2017-03-26
Fixes for Perl6 FOREIGN_LANG

skids:


src/QRegex/Cursor.nqp
```diff
@@ -95,10 +97,10 @@ role NQPCursorRole is export {
 #        nqp::die("No braid in actions!") unless $!braid;
         nqp::atkey(nqp::getattr($!braid, Braid, '$!slangs'),$name ~ "-actions");
     }
-    method define_slang($name,$grammar,$actions = "none") {
+    method define_slang($name,$grammar,$actions = NO_VALUE) {
 #        nqp::die("No braid in define_slang!") unless $!braid;
         nqp::bindkey(nqp::getattr($!braid, Braid, '$!slangs'),$name, $grammar);
-        nqp::bindkey(nqp::getattr($!braid, Braid, '$!slangs'),$name ~ "-actions", $actions) unless nqp::isnull($actions) || $actions eq 'none';
+        nqp::bindkey(nqp::getattr($!braid, Braid, '$!slangs'),$name ~ "-actions", $actions) unless nqp::isnull($actions) || $actions =:= NO_VALUE;
```
zoffixznet: Since it checks for null, wouldn't `nqp::null` default make more sense?

skids:
Along with rakudo PR#1049 this fixes some issues discovered while
getting Grammar::BNF back into working condition on fresh checkouts.

src/QRegex/Cursor.nqp
skids: I don't know why a lot of code NQP code seems to do this.  See e.g. src/QAST/Node.nqp.
So if that's a relic from days past, it should probably be cleaned up tree-wide with a different janitor patch.

zoffixznet: Looking at code in `src/QAST/Node.nqp`, all those `NO_VALUE` cases do not equate no value as being the same thing as `nqp::null`. But in this method we do.

skids: Yeah, I dunno.  All I know is it was throwing warnings the way it was.  Maybe ask TimToady why it was that way in the first place.

zoffixznet: That's 'cause old version uses `$actions` as a string during the test.

skids: Well, yes, of course.  But why it used that string in the first place, I don't know.  Maybe it has something to do with the role being shared between nqp and Perl6 classes?

skids: Seems to spectest fine with just nqp::null at least.  Also I grepped to check that there are no obvious uses of "none" when calling it in rakudo itself.

zoffixznet: Sorry, should this be merged as is or will you submit an update with `nqp::null` as default?

skids: Updated to just use nqp::null

zoffixznet:
Thanks.

PR #350, opened 2017-03-30
[moar] Use nqp::indexic op in QASTRegexCompilier

samcv:
The only remaining places in QASTRegexCompilier for MVM are used for
character ranges, which need to stay lc and uc to make character ranges
make sense.

This takes advantage of the MoarVM case insensitive string index operator,
giving us at least 1.8x speed boost doing case insensitive regex.


samcv:
Merging :-) @zoffixznet++ for reminding me of the PR before our release. Would have been a shame if the case insensitive regex improvements were not included in the release!

PR #351, opened 2017-04-13
Make errors in QASTOperationsMAST more awesome

MasterDuke17:
Make the wording a bit more consistent and include relevant information
where possible. E.g., if the number of arguments to an op is wrong, say
both the number expected and received; if the type of an argument is
wrong, say both the type expected and received.

NQP passes `make m-test` and Rakudo passes `make m-spectest`.


zoffixznet:
Thanks.

PR #352, opened 2017-04-13
Make errors in QASTCompilerMAST more awesome

MasterDuke17:
Make the wording a bit more consistent and include relevant information
where possible. Similar to 513d4eb.

NQP passes `make m-test` and Rakudo passes `make m-spectest`.


zoffixznet:
Thanks.

PR #354, opened 2017-04-22
Make profile sql output better

MasterDuke17:
Change two table names: `allocators` -> `types` and `callees` -> `calls`

Use `0`s instead of `NULL`s when there's no value (unneeded carry-over from a JSON output optimization).

Rename `id` fields that refer to another table's `id` field to `<other table>_id`.

Add the necessary information to the `allocations` table such that all the functionality of the "Allocations" tab in the HTML profile can be replicated.

Change the `caller_id` and `callee_id` in the `calls` table such that it's possible to find out what
routines call a particular routine.

### DIFF
--- a/src/vm/moar/HLL/Backend.nqp
+++ b/src/vm/moar/HLL/Backend.nqp
@@ -308,18 +308,21 @@ class HLL::Backend::MoarVM {
                 }
                 elsif $k eq 'call_graph' {
                     my %callee_rec_depth;
-                    sub collect_callees($caller, %call_graph) {
-                        my @callee := nqp::list_s(~$caller);
+                    my int $node_id := 0;
+                    sub collect_callees(str $caller_id, %call_graph) {
+                        my str $callee_id := ~$node_id;
+                        $node_id++;
+                        my @callee := nqp::list_s($caller_id, $callee_id);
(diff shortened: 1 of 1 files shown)


MasterDuke17:
Sample queries:

- allocations by type: `select t.name, sum(a.count) as s from allocations a, types t where a.type_id = t.id group by t.name order by s desc;`
- allocations by routine for a select type: `select r.name, r.line, r.file, sum(a.count) s from allocations a, types t, calls c, routines r where a.type_id = t.id and t.name = "BOOTHash" and a.callee_id = c.callee_id and c.routine_id = r.id group by r.name, r.line, r.file order by s desc;`
- "Routines" tab output: `select case when r.name = "" then "<anon>" else r.name end, r.file, r.line, sum(entries) as entries, sum(case when rec_depth = 0 then inclusive_time else 0 end) as inclusive_time, sum(exclusive_time) as exclusive_time from calls c, routines r where c.routine_id = r.id group by c.routine_id order by inclusive_time desc;`
- routines that call a specific routine: `select distinct r.name, r.line, r.file from routines r, (select distinct c.caller_id from calls c, routines r where c.routine_id = r.id and r.name = 'CALL-ME') ca, calls c2 where c2.callee_id = ca.caller_id and c2.routine_id = r.id;`

timo:
thanks very much!

PR #357, opened 2017-05-03
Zero slot when popping an array

MasterDuke17:
Otherwise ghost elements are left.

Fixes https://rt.perl.org/Ticket/Display.html?id=131245 for the JVM.

Previously `perl6 -e 'my @a=(1,2,3,4,5); @a.pop; @a.pop, @a[4]="hi"; say @a.join(",");'` would print `1,2,3,4,hi`, now it prints `1,2,3,,hi`.

NQP passes `make j-test` and Rakudo builds and I'm currently running a `make j-spectest`.


zoffixznet:
Thanks.

PR #358, opened 2017-05-06
Align JVM x operator with MoarVM x operator

MasterDuke17:
Do the same error checks (except with different max values depending on
the limitations of the specific VM).

NQP passes `make j-test`. Rakudo passes `make t/spec/S03-operators/repeat.t`.


zoffixznet:
Thanks.

PR #367, opened 2017-08-22
Fix 'Substr length cannot be negative' from Match

ugexe:
I don't understand why this should change anything, but it does.

Fixes Grammar::HTTP - see:
https://github.com/ugexe/Perl6-Grammar--HTTP/issues/6
https://irclog.perlgeek.de/perl6-dev/2017-05-02#i_14521237

### DIFF
--- a/src/QRegex/Cursor.nqp
+++ b/src/QRegex/Cursor.nqp
@@ -87,11 +87,11 @@ role NQPMatchRole is export {
     method to()     { $!to < 0 ?? $!pos !! $!to }
     method CURSOR() { self }
     method PRECURSOR() { self."!cursor_init"(nqp::getattr($!shared, ParseShared, '$!target'), :p($!from)) }
-    method Str()       { $!pos >= $!from ?? nqp::substr(nqp::getattr($!shared, ParseShared, '$!target'), $!from, nqp::sub_i(self.to, $!from)) !! '' }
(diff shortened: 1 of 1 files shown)


ugexe:
Note the difference (current master):
```
$ perl6 -e 'grammar Foo { token TOP { <tox> }; token tox { .+ }; }; say Foo.parse("xyz")'
｢xyz｣
 tox => ｢xyz｣

$ perl6 -e 'grammar Foo { token TOP { <to> }; token to { .+ }; }; say Foo.parse("xyz")'
｢｣
 to => ｢｣
```

This PR fixes the second case.

lizmat:
Wouldn't just replacing `self.do` with `$!do` also fix the issue?

ugexe:
@lizmat Doing so results in `Substring length (-358) cannot be negative` during stage2

lizmat:
Ok, that would imply that the `to` method is not a simple accessor.  Maybe there's more logic in there that you would need to emulate.

ugexe:
That just shows the `to` method logic (which is inlined in this PR) is needed for cases when `$!to` is negative. In the golfed grammar example the `token to {  }` changes the behavior of `NQPMatchRole.to` internally. Yet if we change 2 instances of `self.to` to `$!to < 0 ?? $!pos !! $!to` (the inlined code of `self.to`), as seen in the PR, the `token to { }` no longer affects the internals.

W4anD0eR96:
@ugexe is this problem fix other place?

PR #369, opened 2017-08-29
Do not start interactive REPL on non tty input

andreoss:


andreoss:
This should make handling non tty input in the Rakudo's REPL unnecessary (https://github.com/rakudo/rakudo/commit/2c0cd0a3a96d8fe0b1bb821c33fe51912e95ddfc)

Before 
```
% echo 'say $*PROGRAM-NAME' | perl6
interactive
```
After
```
% echo 'say $*PROGRAM-NAME' | perl6
-
```

lizmat:
Feels to me this should be one `elsif` with `!@a` with an `if stdin().t()`/`else` inside of it, and then lose the `my $interactive`

zoffixznet:
👍 Thanks!

> Feels to me this should be one elsif with !@a with an if stdin().t()/else inside of it, and then lose the my $interactive

Agreed. I'll make that change before version-bumping.

PR #370, opened 2017-09-06
Map the new nativecallinvokejit OP

niner:
nativecallbuild now has a return value indicating whether we were able
to JIT compile code for the call site


niner:
This has long been merged

PR #373, opened 2017-10-11
[jvm] Don't wait for child when program ends

usev6:
With a user thread the following program sleeps for another 18 seconds
after saying bye:

  my $proc = Proc::Async.new('sleep', '20');
  my $promise = $proc.start;
  sleep 2;
  say 'bye';

With a daemon thread the program exits after saying bye. The child
process runs for another 18 seconds on its own. That's the behaviour
with MoarVM as well.


jnthn:
Thanks!

PR #381, opened 2017-11-08
Add getppid op for moarvm

ugexe:
Allows a child process to get the parent process id, and thus guess if it has become orphaned.

### DIFF
--- a/docs/ops.markdown
+++ b/docs/ops.markdown
@@ -297,6 +297,7 @@
   * [getenvhash](#getenvhash)
   * [backendconfig](#backendconfig)
   * [getpid](#getpid)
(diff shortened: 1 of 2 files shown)


ugexe:
Requires: https://github.com/MoarVM/MoarVM/pull/748

[This node PR](https://github.com/nodejs/node/pull/16839) should make the js implementation simple enough once it lands.

[Java 9](https://docs.oracle.com/javase/9/docs/api/java/lang/ProcessHandle.html#parent--) is adding a Process/ProcessHandle interface that should make this trivial (as well as an alternative to the jvm backends current implementation of `getpid`), but not sure about Java 8

PR #383, opened 2017-12-07
Fix op_if/unless/with/without coercion

W4anD0eR96:
Sample code:
```
my $foo := 42;
my int $bar := $foo || 200;
say($bar);
```

Before it says '0' and now '42'.
The problem is we generate $foo || 200 as

```
- QAST::Op(unless &infix:<||>)  ||
  - QAST::Var(lexical $bar)
  - QAST::IVal(200)
```
and thus $bar only be a condition. If we want to return its value,
we forget to coerce its kind.

Now this fix only apply on op_unless. If applies on the four ops
handled this block, it fails some tests in t/qast/01-qast.t.
For details
```
t/qast/01-qast.t                     (Wstat: 0 Tests: 166 Failed: 3)
  Failed tests:  156, 162, 165
```


W4anD0eR96:
No, this break 

```
my $foo := 0;
my int $bar := $foo || 400;
say($bar);
```

to say 0. I will work on it.

W4anD0eR96:
Now we generate correct condition IR, it's quite like a copy from JVM backend.

```
a || b

ifFalse a goto L1
then-clause
goto L2
L1:
{ coerce if oprands == 2 and needed }
else-clause
L2:
...
```

PR #386, opened 2018-01-06
[jvm] Look inside containers for parametric type

usev6:
Fixes some spectests (cmp. RT #132673).


usev6:
I'm not sure if this patch make sense -- but it seems to fix some roast issues on the JVM backend.

jnthn:
I think `decont` itself does the check of whether it needs to, so the `iscont` check here is perhaps duplicated work?

usev6:
Of course, you are right. I didn't realize that decont is a noop if it gets a decontainerized argument. Thanks, jnthn!

If the addition of decont makes sense otherwise, I could as well push that change directely (avoiding the intermediate and the merge commit).

jnthn:
Yes, it makes sense to me.

usev6:
I commited the patch as f6bccf79d8 and I'm closing this PR. Thanks again!

PR #387, opened 2018-01-08
Rakudo Issue 1304: Preserve operator chaining when operators negated

jstuder-gh:
This is one part in a two part fix to [Rakudo Issue 1304](https://github.com/rakudo/rakudo/issues/1304).
The Rakudo component to this fix can be found [here](https://github.com/rakudo/rakudo/pull/1375).

Modify the 'chain' op to allow the option to use the first child as the
callee. Before the name of the op served as the operator sub and the
children the operands. This modification makes it so that, if there is
no name provided to the chain, child 0 serves as the operator and
children 1 and 2 the operands.

allowing negated chained ops to continue to work as chained.


zoffixznet:
👍 Thanks!

PR #390, opened 2018-01-17
Rakudo Issue 1304: [js] Allow 'chain' op to use child as callee

jstuder-gh:
This is a modification to the js compiler allowing the 'chain' op to use the
first child as callee if the name of the op is not set.

(from a little while back) allowing negated chained ops to continue to work as chained.

See <https://github.com/rakudo/rakudo/issues/1304>.

### DIFF
--- a/src/vm/js/Operations.nqp
+++ b/src/vm/js/Operations.nqp
@@ -258,18 +258,25 @@ class QAST::OperationsJS {
 
         my sub chain_part($part) {
             if is_chain($part) {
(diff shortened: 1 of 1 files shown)


jstuder-gh:
Note that I tested this on a branch where I cherry-picked the commit [ba111cfed474e6c7091b84dab5568f5174262884](https://github.com/rakudo/rakudo/commit/ba111cfed474e6c7091b84dab5568f5174262884) on top of the js branch in Rakudo. So while not a part of the js branch yet, once the js branch incorporates that commit chaining of negated ops should be ready to go.

pmurias:
Thank you

PR #391, opened 2018-01-17
remove unreachable code in cursor_init

dwarring:
we're already in an else block from a previous
nqp::isconcrete(self) && $!braid test


lizmat:
Thanks!

PR #392, opened 2018-01-20
change .markdown file suffix to .md

tbrowder:
rename markdown files to a shorter and more popular convention


zoffixznet:
👎 on this. 

I personally linked to the docs in many places many times. I'm sure others have done the same. Renaming the files would result in all those links becoming 404s.

AlexDaniel:
Yes, I agree. This is unfortunate but we'll have to reject it. This concern applies to many other renames (e.g. ``pl`` →``p6``), and while it's sad we can't be breaking links all the time.

If you have any ideas on how to improve the situation, then we can definitely discuss it. For example, I'm not a fan of storing the documentation in a git repository. IMO it makes micro-contributions less likely (because even if it is relatively easy to change files via github interface, no one wants to cause git history noise by changing one letter somewhere). I don't know if there's a better approach to manage nqp docs, but I'm ready to hear the ideas.

Anyway, `.markdown` is not too bad, so unless it causes some technical issue (e.g. tools not understanding the extension), let's not rename.

AlexDaniel:
By the way, github wiki is *not* a solution, because [redirects are not supported](https://github.com/gollum/gollum/issues/1023). Eh.

PR #394, opened 2018-01-20
add new pod tests

tbrowder:
+ test for robust =end for blocks even though misaligned with its =begin

+ test to show current state of failure with nested comment blocks
  that shows parsing of code ends at the first =end comment
  OR an error is thrown in other circumstances


lizmat:
Thanks!

PR #395, opened 2018-01-21
Don't die if QAST parent doesn't have dump_extra_node_info method

jstuder-gh:
The dump_extra_node_info method of the QAST::SpecialArg role would look
up the method of the same name on the immediate parent and invoke it,
dying if that method didn't exist and could not be invoked (like when mixed-in
to QAST::Want, for instance). Modifying the behavior so that it checks
whether the there is an invokable method, and if not continue along the
parent chain until one is found or all are exhausted.

### DIFF
--- a/src/QAST/SpecialArg.nqp
+++ b/src/QAST/SpecialArg.nqp
@@ -15,8 +15,18 @@ role QAST::SpecialArg {
     }
 
     method dump_extra_node_info() {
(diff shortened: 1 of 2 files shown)


jstuder-gh:
I don't know if this is the best place to put the accompanying test. If it is not, let me know and I'll move it elsewhere.

lizmat:
Looks good to me in casual inspection.

PR #397, opened 2018-01-23
convert tabs to 4 spaces, correct some obvious code misalignment

tbrowder:


src/QRegex/P6Regex/Actions.nqp
```diff
@@ -631,11 +631,11 @@ class QRegex::P6Regex::Actions is HLL::Actions {
         my int $i := 1;
         my int $n := nqp::elems($clist);
         while $i < $n {
```
AlexDaniel: the misalignment here is broken now?

AlexDaniel:
I support the changes, but this PR breaks some parts. Github diffs shows that crearly.

zoffixznet:
Looking at the diff on GitHub, looks like some parts become messed up.

I'm also -½ on such sweeping changes because they muddy up `git blame`. I rather any such fixes get applied whenever that specific part of code needs changes.

z

zoffixznet:
> because they muddy up git blame

Example of what I mean. Here's a chunk of code got broken up and reader needs to dig one commit back to skip over alignment commit to see the commit for the code:

z

tbrowder:
Shall I close this and submit one at a time?  Or correct each somehow to make the blame diffs look right.  Or.

(As an aside, I would also like to remove cuddled elses.)

zoffixznet:
> Shall I close this and submit one at a time (or maybe one line change at a time)? Or correct each somehow to make the blame diffs look right. Or?

I meant this would be corrected whenever that code needs to be changed for some other reason, so the blame would just say "fixed such and such" instead of formatting changes. But I don't feel strongly about this, so 🤷‍♀️

AlexDaniel:
> Shall I close this

As for fixing the issues that I mentioned, no, just `git commit --amend` it.

As for git blame, there's nothing you can do on your side. Those who use git blame can use ``-w`` (``-M`` is also a useful one) to ignore whitespace changes. Github should support ``?w=1`` parameter on blames (it already does for diffs), maybe we should report that somewhere (github doesn't have a bug tracker iirc, oh the irony…).

> I meant this would be corrected whenever that code needs to be changed for some other reason

In practice the code gets corrected when some code around (but still far away) is changed, so in the end you still have to do the blame jump, but now with a commit message that is irrelevant. IMO changing all at once is much better, especially considering that a lot of things won't be touched in a while and living with awful indentation for that time for me is unjustified.

tbrowder:
ok

tbrowder:
well, i probably messed up that git --amend, but git got in my way.  anyway, the change is there.  the only way i know to clean it up is to close this and redo. easy enough to do.

tbrowder:
I’m closing this and will resubmit a new PR.

PR #400, opened 2018-01-25
add new test file for nqp using perl

tbrowder:
+ provides a test harness for testing known failures
+ test for known failure using =cut

### DIFF
--- a/tools/build/Makefile-JVM.in
+++ b/tools/build/Makefile-JVM.in
@@ -302,6 +302,7 @@ j-bootstrap-files: $(J_STAGE2_OUTPUT)
 
 j-test: j-all
 	$(PROVE) -r --exec .@slash@$(J_RUNNER) t/nqp t/hll t/qregex t/p5regex t/qast t/serialization t/jvm t/nativecall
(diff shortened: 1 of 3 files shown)


zoffixznet:
`xt/` is generally for author tests while `t/` is for user tests. This PR seems to be adding POD tests as author tests (is that necessary?) while at the same time it's adding author tests to be run with `make test`, along side user tests.

tbrowder:
The test should be standard, but they use perl 5 so they need to be in a separate dir from the “exec nqp” tests.  I forgot the special convention for xt, how about some other, unique dir then like “t/perl5/<lang>”?

zoffixznet:
Personally, I'd just write the nqp version of STDERR capturing routine instead of writing an entire test file in Perl 5 just to execute a program that creates a file.

JS back end even [already has an implementation that can be adapted](https://github.com/perl6/nqp/blob/b88de49aadefc1fe758197d51a71cc8173064dbf/src/vm/js/HLL/Backend.nqp#L4-L69)

tbrowder:
I’ll close this and resubmit after we decide where to put the Perl 5 tests for <lang>.

PR #402, opened 2018-01-27
Rakudo Issue 1304: [jvm] Allow 'chain' op to use child as callee

jstuder-gh:
This is a modification to the jvm compiler allowing the 'chain' op to use the
first child as callee if the name of the op is not set.

(from a little while back) allowing negated chained ops to continue to work as chained.

See <https://github.com/rakudo/rakudo/issues/1304>.


zoffixznet:
👍 Thanks!

PR #408, opened 2018-02-12
add a test for pod panic, but stderr is blocking

tbrowder:
Look at the new test to see that the general format of the code works for no capture and capture of stdout, but hangs for anything else. Also notice that the existing tests do NOT test for stderr in the 111*async*t file.

### DIFF
--- a/t/nqp/005-pod-panic.t
+++ b/t/nqp/005-pod-panic.t
@@ -49,8 +49,9 @@ sub run-command($command, :$stdout, :$stderr) {
     my @out-bytes;
     my $queue := nqp::create(Queue);
 
-    my $done     := 0;
-    my $read-all := 0;
+    my $done         := 0;
+    my $read-all     := 0;
+    my $calles-ready :=0;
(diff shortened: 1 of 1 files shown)


t/nqp/005-pod-panic.t
```diff
@@ -0,0 +1,146 @@
...
+    }
+    else {
+        ++$read-all;
+    }
+
+    my $task := nqp::spawnprocasync($queue, $command, nqp::cwd(), nqp::getenvhash(), $config);
+
+    if $stdout || $stderr {
+        nqp::permit($task, 1, -1);
+    }
```
jnthn: Pretty sure this is wrong; for the `$stderr` only case, it would need to be `2` here, not `1`.

tbrowder: ok, a good hint to check! thanks!

tbrowder:
closing for now while investigating @jnthn’s suggestion.

PR #411, opened 2018-02-13

tbrowder:
  using the new sub run-command

### DIFF
--- a/src/core/IO.nqp
+++ b/src/core/IO.nqp
@@ -293,4 +293,145 @@ sub join($delim, @things) {
     nqp::join($delim, @strs)
 }
 
+sub run-command($command, :$stdout, :$stderr) {
+    my class Queue is repr('ConcBlockingQueue') { }
+
+    # we alway capture stdout and stderr output, but we only return what is asked for
+    # use one output stream, both, or none
+    # return an array of lines for each stream called for
+    my $line-length := 85;
+    my @stdout-bytes;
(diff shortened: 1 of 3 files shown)


niner:
What's the purpose of this new sub?
Shouldn't it be part of the 6.d setting? At least for subs we can easily stay 100 % backwards compatible and we should use that advantage.

niner:
Ah, sorry, shouldn't have commented before finishing my coffee. Didn't notice that this is the nqp repo.

AlexDaniel:
But even if so, to me it feels like this should be a helper sub for tests only. But maybe I'm missing something.

zoffixznet:
FWIW, there's a [`src/core/testing.nqp` file](https://github.com/perl6/nqp/blob/master/src/core/testing.nqp] for testing routines, though it'd be available in the setting just the same as `is` or `ok` are.

> this should be a helper sub for tests only. But maybe I'm missing something.

I think there's a limitation that NQP modules must be already precompiled, so you have a choice of shoving this helper routine into the setting, repeating it in each test file, or figuring out how to precompile a helper test module.

tbrowder:
I need to be able to test various pod failures in nqp on my road to a slang for pod in nqp and then Perl 6. I do not want to copy/paste so much boiler plate every time I need a new test.

tbrowder:
failing on travis, closing and will reopen after fixing

Issue #409, opened 2018-02-12
spawnprocasync needs tests for stderr and merge captures


tbrowder:
add them to file 111-spawnprocsync.t

MasterDuke17:
@tbrowder can this issue be closed now?

TisonKun:
FYI ref #477 

approve to close this

PR #414, opened 2018-02-14
add a test for stderr in spawnprocasync

tbrowder:
+ provides a partial fix for nqp GH #409 (no test for merge yet)
+ note the test for stderr on moarvm is good, but it is skipped for the
  jvm because something appears to be wrong with the stderr stream
  interface there
+ adds comments for the tests
+ organizes the code to be more readable

### DIFF
--- a/t/nqp/111-spawnprocasync.t
+++ b/t/nqp/111-spawnprocasync.t
@@ -54,7 +54,7 @@ my $config := nqp::hash(
 );
 
 # define a task
-$string  := "expected output";
+$string  := "expected output from stdout";
 $command := "echo $string";
 $args := $is-windows ?? nqp::list(nqp::getenvhash()<ComSpec>, '/c', $command)
                      !! nqp::list('/bin/sh', '-c', $command);
@@ -65,11 +65,11 @@ nqp::permit($task, 1, -1);
 while !$done || !$read_all {
   if nqp::shift($queue) -> $task {
     if nqp::list($task) {
-      my $code := nqp::shift($task);
-      $code(|$task);
+        my $code := nqp::shift($task);
+        $code(|$task);
     }
     else {
-      $task();
+        $task();
     }
   }
 }
@@ -89,7 +89,7 @@ $read_all     := 0;
 $called_ready := 0;
 
 # define the task
-$string  := "expected output";
+$string  := "expected output from stderr";
 $command := "echo $string >&2";
 $args := $is-windows ?? nqp::list(nqp::getenvhash()<ComSpec>, '/c', $command)
                      !! nqp::list('/bin/sh', '-c', $command);
@@ -100,11 +100,11 @@ nqp::permit($task, 2, -1);
 while !$done || !$read_all {
   if nqp::shift($queue) -> $task {
     if nqp::list($task) {
-      my $code := nqp::shift($task);
-      $code(|$task);
+        my $code := nqp::shift($task);
+        $code(|$task);
     }
     else {
-      $task();
+        $task();
     }
   }
 }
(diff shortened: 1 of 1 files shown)


usev6:
I'm afraid, this test assumes too much about the output of '/bin/sh -c foo':

```
debian7$ /bin/sh -c foo
/bin/sh: 1: foo: not found
```

```
freebsd11$  /bin/sh -c foo
/bin/sh: foo: not found
```

There is no line number on the second host, so the test fails.

Apart from that (if I adjust the regex a bit) the test *does* on JVM for me. I don't see offhand why the last version (PR 412) failed on travis. I'll try to take a look (compiling nqp-j on debian now).

tbrowder:
Good call, I can definitely make the regex looser, didn’t think about other *nixes.

The previous PR failed because a bad test was on the master branch and I didn’t test the master branch before I started the previous branch.

Do you know where we can get access to other *nixes to test these kinds of things? I guess I can create some vms on Digital Ocean or other cood services to do that.

usev6:
> Do you know where we can get access to other *nixes to test these kinds of things? I guess I can create some vms on Digital Ocean or other cood services to do that.

Sorry, I don't have a good idea here.

But maybe we can find a better (more robust) way to test for stderr? I'm not sure, but I've heard that 'echo' and output redirection to stderr works on *nix as well as on Windows. So maybe this would work:

```
# define the task
my $command := 'echo string_that_should_appear_on_stderr >&2';
$args := $is-windows ?? nqp::list(nqp::getenvhash()<ComSpec>, '/c', $command)
                     !! nqp::list('/bin/sh', '-c', $command);
[...]
is(nqp::decodertakeallchars($dec), "string_that_should_appear_on_stderr\n", 'got the correct output on stderr');
```

In roast we have something similiar: https://github.com/perl6/roast/blob/f2d745f1a80fdb00a012c1c379921e4efdd14bc6/S29-os/system.t#L53

Another note regarding the JVM: The test failed for me on Linux, too. So it looks like there are some issues with capturing spawnprocasync and capturing stderr and it's fine to skip the test on JVM for now.

tbrowder:
Ok, @usev6, good suggestions.  I'll keep this open for now is case you have some more ideas, but I'll try your current suggestions and, if I can get good results, I'll close this and resubmit with a new PR.

Do you think I should file the issue of the stderr problem on the jvm?  I do, especially since you see the same thing.

niner:
Even a looser regex would fail on this one:

nine@sphinx:~> /bin/sh -c foo
/bin/sh: foo: Kommando nicht gefunden.

tbrowder:
Ok, @usev6’s suggestions will definitely be investigated. No need t go down that multilanguage OS rabbit hole! Thanks, @niner.

t/nqp/111-spawnprocasync.t
```diff
@@ -37,32 +42,83 @@ my $config := nqp::hash(
...
+$string  := "expected output";
+$command := "echo $string >&2";
+$args := $is-windows ?? nqp::list(nqp::getenvhash()<ComSpec>, '/c', $command)
+                     !! nqp::list('/bin/sh', '-c', $command);
+$task := nqp::spawnprocasync($queue, $args, nqp::cwd(), nqp::getenvhash(), $config);
+nqp::permit($task, 2, -1);
+
+# run the task
+while !$done || !$read_all {
+  if nqp::shift($queue) -> $task {
```
AlexDaniel: This block seems to have 2 space indentation while everything else is using 4.

tbrowder: Ok, will fix.

tbrowder: Changes made and committed, all tests passing.

PR #419, opened 2018-02-18
Fix eof detection on zero-size files

usev6:
Should help with https://github.com/rakudo/rakudo/issues/1541

There is still an issue with failing spectest for eof on Rakudo that use .read, but I think those are caused by something else.


usev6:
Travis failure was caused be wrong test count -- already fixed with 350f4b5743

zoffixznet:
Ah, now I see why my original code wasn't working :) 👍 Thanks!

PR #421, opened 2018-02-21

tbrowder:
  using the new sub run-command


tbrowder:
argh! not the right code

PR #424, opened 2018-02-22

tbrowder:
  using the new sub run-command
+ fix test for spawnprocasync
+ document all testing routines in "built-ins.md"


tbrowder:
A major finding was the 10+ minutes for testing stalled travis-ci, even though tests were good.  The mods to .travis.yml seemed to fix that.

An unsolved problem still results in t/nqp/114*t working on the jvm locally on my host, but not on travis-ci jvm, thus that test is skipped for the jvm.  Since stderr is being used it may be affected by the PR to fix eof handling (by @pmurias).

PR #425, opened 2018-02-22
[moar] Add encodeconf and decodeconf ops

samcv:
These ops by default decode strictly with windows-1252 and windows-1251
codetable specification, throwing on invalid input or output (as is done
now for utf8 encoding). They can be set to decode/encode permissively as
well, which will decode/encode codepoints that are unmapped yet still
fit within one byte.


tbrowder:
@samcv, if you make your .travis.yml file the same as on the master branch, that should cure the job timeout problem with the jvm travis run.

W4anD0eR96:
close via https://github.com/perl6/nqp/commit/b6d98eac3f0dff1dbd4df252351efa35ab700f7e

PR #429, opened 2018-03-04
[JVM] Let seekfh return offset, not file handle

usev6:


usev6:
Closed in favour of 46fbb5e9d8

PR #430, opened 2018-03-08
Rakudo Issue 1588: Create getsignals op for MoarVM / JVM

jstuder-gh:
This op will retrieve the signals available on the host system. Also
removing the signal-related nqp const values as they are no long valid
with this change.

See <https://github.com/rakudo/rakudo/issues/1588>


jstuder-gh:
Corresponding Rakudo PR: <https://github.com/rakudo/rakudo/pull/1601>

jstuder-gh:
Relies on PR <https://github.com/MoarVM/MoarVM/pull/814>

jstuder-gh:
I'm hoping to bring attention back to this PR (and its corresponding [Rakudo component](https://github.com/rakudo/rakudo/pull/1601)) as it's been in the queue for a while and may have been forgotten. However, I still feel that it is useful and solves a [real issue](https://github.com/rakudo/rakudo/issues/1588).

What this op does is provide a hash of the signals that are available on the system, which ultimately, can be used to [populate the Signal enum values](https://github.com/jstuder-gh/rakudo/blob/774ec5a0cb3e00fa4f5159228f540632277f2d30/src/core/signals.pm6#L1). Signals that are not available have a value of 0 and signals that are not supported by the backend being used are negative.

The MoarVM op has been merged, so now this PR and the Rakudo one need merging in order to complete the issue for the that backend. The Travis tests that failed when I opened this PR would succeed now.

Since the JVM implementation of the slice op wasn't without issue, I'm hoping for a review of the JVM implementation of this op from an expert. @usev6, when you get the time could you please take a look and let me know if there is anything I can do better?

jstuder-gh:
I've just added a getsignals op for the JS backend as well.

Also, I've updated the JVM implementation to address an issue I noticed recently. See the commit message for e9c9ca7f515d44ac86f201f26c3b900f6ddf7624 for more details, but I think this is a positive change. I've also broken up the logic previously within the getsignals method into several methods within a SigProcess subclass, which I also believe to be helpful.

jstuder-gh:
Also wanted to point out that with the new commits all Travis builds pass. The previous failures were due to the fact that, at the time it was originally submitted, the MoarVM getsignals op did not exist. The MVM PR has since been merged and now these tests pass as a result.

PR #432, opened 2018-03-23
Rakudo Issue 1622: Fix bug with iteration of empty character class

jstuder-gh:
Check that there is at least one character class before applying the
regex nodes.

When compiling the 'conj' regex node, it expects to be able to shift
off at least one child node and iterates past the end as a result.

Fixes [Rakudo Issue #1622](https://github.com/rakudo/rakudo/issues/1622)

### DIFF
--- a/src/QRegex/P6Regex/Actions.nqp
+++ b/src/QRegex/P6Regex/Actions.nqp
@@ -765,8 +765,8 @@ class QRegex::P6Regex::Actions is HLL::Actions {
             @alts.push(QAST::Regex.new( $str, :rxtype<enumcharlist>, :node($/), :negate( $<sign> eq '-' ),
                                         :subtype($RXm ?? 'ignoremark' !! '') ))
                 if nqp::chars($str);
-            $qast := +@alts == 1 ?? @alts[0] !!
-                $<sign> eq '-' ??
+            $qast := ( my $num := +@alts ) == 1 ?? @alts[0] !!
+                0 < $num && $<sign> eq '-' ??
                     QAST::Regex.new( :rxtype<concat>, :node($/), :negate(1),
                         QAST::Regex.new( :rxtype<conj>, :subtype<zerowidth>, |@alts ),
                         QAST::Regex.new( :rxtype<cclass>, :name<.> ) ) !!


lizmat:
Looks good to me.

jstuder-gh:
Ok then if there are no objections I will merge this in tonight.

jnthn:
I think this should probably give an error rather than work; surely nobody ever intends to write an empty character class.

AlexDaniel:
But ``<[]>`` does not give an error, so this PR seems to be consistent.

jstuder-gh:
I experimented with adding an error message like so ([see this commit](https://github.com/jstuder-gh/nqp/commit/7c41dfa72a4e7b4f21e148b990d3b11e382e6f0e)), but then NQP would not build without displaying that error so I backed off from that. 

I can try experimenting a bit more with it tonight, but that was the reason I submitted it without the error. Looking at it again, that isn't the best place for that assertion (as it's only checking the first character class), so maybe it could work but is just poorly placed.

jstuder-gh:
I've updated the branch so that it will panic if it encounters an empty char class (<[ ]>, <+[ ]>, or <-[ ]>). I had to add the check for whether the rxtype is 'cclass'; cclass_backslash nodes have no children and were causing the NQP build to fail. 

When running the spectest [these tests](https://github.com/perl6/roast/blob/master/S05-mass/stdrules.t#L313) cause it to die and thus fail. Do we modify the tests in this case?

jstuder-gh:
Looking at the [RT thread](https://rt.perl.org/Public/Bug/Display.html?id=109674) related to the [tests that fail](https://github.com/perl6/roast/blob/master/S05-mass/stdrules.t#L313) with the panic added, it looks like the empty character class is an intentional thing. I guess I'll remove the panic commit? That looks to be the proper behavior.

AlexDaniel:
These tests are part of 6.c-errata. I can't say if it's a “proper” behavior or not, but it is what it is in v6.c. We can easily merge this without the panic commit, or we can merge it if it acts this way in v6.d only.

Well, unless someone can justify why the current behavior is very wrong.

jstuder-gh:
How about I remove the panic commit from this branch, reapply on another branch, and submit a separate PR or issue inviting comment on whether the panic is appropriate to include in v6.d? That way we can merge in the sure thing and defer the panic question for now.

Can we specify in the roast that certain blocks adhere to a particular standard (ie lexically scoped 'use v6.*' statements)? So if we add a test to indicate that the <-[ ]> does not error now, and if the panic behavior is adopted for v6.d we can clarify the old tests apply to v6.c and make new tests that apply only to v6.d?

AlexDaniel:
@jstuder-gh Your patch should work both under v6.c and v6.d at the same time, doing the “right” thing depending on which version is used (e.g. `use v6.c` vs `use v6.d.PREVIEW`).

Yes, please split it in two and we'll discuss the second part separately.

jstuder-gh:
@AlexDaniel I re-pushed the branch but this time without the panic commit. I will resubmit that commit as a separate PR soon.

PR #435, opened 2018-04-07
[RFC] Panic when given empty character class <[ ]>

jstuder-gh:
Description: 
Check whether the QAST node has any children to determine if it is
empty.  The exception are the backslash cclass nodes (\b, \n, \s, etc).

When providing a fix for [Rakudo Issue 1622](https://github.com/rakudo/rakudo/issues/1622), there was some discussion about whether the empty character class should be allowed at all ([see discussion here](https://github.com/perl6/nqp/pull/432)).

Should this panic behavior be adopted as part of the official behavior of Perl 6 in future releases (6.d+) despite it being somewhat at odds with the behavior detailed in the roast before?

### DIFF
--- a/src/QRegex/P6Regex/Actions.nqp
+++ b/src/QRegex/P6Regex/Actions.nqp
@@ -772,6 +772,13 @@ class QRegex::P6Regex::Actions is HLL::Actions {
                         QAST::Regex.new( :rxtype<cclass>, :name<.> ) ) !!
                     QAST::Regex.new( :rxtype<alt>, |@alts );
         }
+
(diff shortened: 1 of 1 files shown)


JJ:
Is there any point in keeping these PRs here? Can they be salvaged/closed? First thing would be; I guess, to rebase to current main branch, right?

lizmat:
Good question. I propose we look at all of the PRs in NQP / Rakudo again after RakuAST is fully integrated.  It's too much of a shifting landscape at the moment, I'd say.

PR #436, opened 2018-04-20
Expose 'slice' op in NQP (MoarVM / JVM/ JS)

jstuder-gh:
Exposes a lower-level means of taking positional slices.

    slice(@arr, int $start_elem, int $end_elem --> @copy)

Requires this [MoarVM PR](https://github.com/MoarVM/MoarVM/pull/843)

### DIFF
--- a/docs/ops.markdown
+++ b/docs/ops.markdown
@@ -781,9 +781,19 @@ to.
 * `slice(@arr, int $start_pos, int $end_pos --> @copy)`
 
 Copy the elements in `@arr` starting at `$start_pos` and ending at `$end_pos`
-and return the resulting list.  If `$start_pos` is a negative value, the start
-value is set to position 0. If `$end_pos` is a negative value, the end value is
-set to the position of the last element in the list.
+and return the resulting list. If `$start_pos` or `$end_pos` is ```-n``` it will
+translate into the ```n```th position relative to the end of the list.
+
+```perl6
+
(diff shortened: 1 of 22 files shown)


jstuder-gh:
Added additional commits. In addition to exposing the MoarVM slice op, also added commits to port slice to the P6 runtime for the JVM backend.

Changed the name of the PR to indicate these additions.

lizmat:
I'm not sure I like the semantics of negative indices.  In things like `atpos`, a negative index means substracted from `nqp::elems`, so `-2` would be the same as `nqp::elems - 2`

$ 6 'use nqp; my $l := nqp::list(1,2,3); say nqp::atpos($l, -2)'
2

It would seem to me that the same semantics would be expected / useful for the `slice` op.

And apologies for not digging into this sooner.  I *really* like that you're doing this!

jstuder-gh:
You're right, that would be better. Not only is it more consistent but it also provides more flexibility. I'll modify it to address this issue.

lizmat:
@jstuder-gh looks ok to me, except for the Travis problem: do you have any idea what's going on there?

ugexe:
It would need the appropriate MoarVM PR merged to work

jstuder-gh:
Yeah its kind of weird when a project is split up over several repos. I guess if we had MoarVM as a submodule it could work.

PR #442, opened 2018-05-12
[jvm] Add getppid op and upgrade to jdk9

ugexe:
Note: Will require updating the rakudo travis config to use jdk 9 similar to the one included in this PR

JDK 9 adds a new process api:
https://docs.oracle.com/javase/9/docs/api/java/lang/ProcessHandle.html

* Renamed existing `ProcessHandle` class because I wasn't sure how it
would interact with the new java.lang.ProcessHandle.

* Adds `getppid` op on jvm using the new process api.

* Reimplements `getpid` using the new process api.

`ProcessHandle.Info` could probably be used to implement the cpu time instead of `.getOperatingSystemMXBean()`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ addons:
     - gcc-6
 
 script:
-    - if [[ $TRAVIS_OS_NAME != osx && $NQP_OPTIONS =~ jvm ]]; then sudo -E apt-get install oracle-java9-installer; fi
+    - if [[ $TRAVIS_OS_NAME != osx && $NQP_OPTIONS =~ jvm ]]; then sudo -E apt-get install openjdk-9-jdk; fi
     - perl Configure.pl $NQP_OPTIONS
     - make test
     - make install
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,30 @@
-language: "java"
-    
+language: perl
+
+os:
+    - linux
+    - osx
+
 sudo: false
-dist: trusty
+
+fast_finish: true
 
 install: "echo"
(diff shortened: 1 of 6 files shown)


usev6:
Sorry, I didn't have enough time to look at this, yet. At a first glance your changes are looking good.

Some general thoughts on the question of supported Java version(s): My initial thought was, that we should support both, Java 8 *and* Java 9. But I don't know how to do that. I googled a bit the other day, but didn't find an obvious solution. Also, since there are some major bugs with rakudo-jvm (so that very likely no-one uses it for real things), we could as well leave the old versions behind and try to run on the latest version. (Which seems to be Java 10 by now. The end of public updates for Java 9 has been march 2018 according to http://www.oracle.com/technetwork/java/javase/eol-135779.html)

We'll have to look at Rakudo's code base, too. If I'm not mistaken, it also uses a few no longer supported Modules. (There is the use of sun.misc.Unsafe in src/vm/jvm/runtime/org/perl6/rakudo/RakudoContainerSpec.java -- may it can be replaced by using jdk.unsupported?)

lizmat:
FWIW, at this stage I would go for supporting from Java 10 upward.

coke:
If we had a wide install base of Java, I'd be much more concerned about the loss of Java 8 support, but I think it's ok to bump our requirement to 9.

+1

usev6:
A year has passed and Java 9 and Java 10 are things of the past. A few weeks ago I saw that FreeBSD now has openjdk11 and openjdk12 ports. For me that was the last remaining argument against leaving Java 8 behind. @ugexe Thanks for this PR and your patience!

PR #447, opened 2018-05-21
Fix STDIN error for non-interactive REPL mode

uzluisf:
Do not close STDIN it is the source of the program

Fixes https://github.com/rakudo/rakudo/issues/1824


AlexDaniel:
Any tests for this?

Issue #448, opened 2018-05-21
An AST dump of a sub or regex coerces to string


mryan:
When using --target=ast on code that has a sub or regex you get warning on STDERR like:
Sub object coerced to string (please use .gist or .perl to do that)
OR
Regex object coerced to string (please use .gist or .perl to do that)

MasterDuke17:
@mryan can this issue be closed now?

PR #451, opened 2018-05-23
Mryan ast debug

mryan:
Hello,

I've been (slowly) working on issue #1858 of the doc repository.  Before documenting it, I need to figure out how to use it.  Along the way, I developed three "QAST debugging aids" for myself, one of which I think *may* be useful generally and is the subject of this pull request.

The first commit fixes a minor bug which I created issue #448 to cover.
The next three commits add "guide lines" to highlight the "reach" of blocks when using --target=ast.  These three commits constitute the "maybe generally useful" debugging aid for you to consider.
The last commit adds a patch file for the second of the debugging aids which supresses the printing of some QAST nodes under QAST::Want nodes.  This is a bit "heavy handed" so I've put it in the tools directory just so those who are curious can apply it (using "git am") and give me any feedback you have.

Finally, I'm happy to take any feedback on a "meta" level on whether I should have suggested any of this..

Regards

### DIFF
--- a/src/QAST/Children.nqp
+++ b/src/QAST/Children.nqp
@@ -18,12 +18,16 @@ role QAST::Children {
     }
 
     method dump_node_list(int $indent, @onto, @node_list) {
+        # Guide lines are triggered from here - if this is a Block
+        # node, then we request a guide line next to its children
+        my $guide-requested := nqp::istype(self, QAST::Block) ;
+
         for @node_list {
             if nqp::istype($_, QAST::Node) {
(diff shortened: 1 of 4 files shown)


lizmat:
Thank you for your work!  My first impression is: Great!

However, for judging applicability, it may be useful to actually split up this PR into multiple PR's each with their own "functionality".

e.g. Fix stringification of Sub appears to be unconnected to the other commits.

Making several smaller PR's makes it easier to merge the simple, no-brainers, while allowing to focus more on the PR's that require more thought!

Keep up the good work!

PR #452, opened 2018-05-23
Fix stringification of Sub when using --target=ast

mryan:
Ref: issue #448

### DIFF
--- a/src/QAST/Children.nqp
+++ b/src/QAST/Children.nqp
@@ -25,7 +25,11 @@ role QAST::Children {
             else {
                 nqp::push(@onto, nqp::x(' ', $indent));
                 nqp::push(@onto, '- ');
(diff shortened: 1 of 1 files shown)


lizmat:
Cool!

OOC, have you considered putting the ternary test *inside* the `nqp::push`?  Or do you have a stylistic reason for repeating the `nqp::push(@onto...)`

mryan:
No, I have no stylistic reason - if there's something gained by putting the test inside the push I'm happy to do that.

lizmat:
Well, it would result in a smaller binary, that's all, as there would be only one `nqp::push` op instead of 3.  Also, I'm a firm believer in DRY (as in Don't Repeat Yourself).

So in the end, it's not how I would have done it.  The performance gains are most likely minimal.

PR #453, opened 2018-05-23
Mryan ast debug

mryan:
All changes are to code that dumps QAST nodes when the user specifies "--target=ast" to the command line.  There is no change to anything else.

The first two commits factor out the production of "indent strings" - series of spaces - to a single method in QAST::Node called dump_indent_string.  After these two commits, everything works exactly the same as before including when "--target=ast" is specified.

The third commit modifies the dump_indent_string method to take a named parameter,  :$guide-line  which produces, well, a guide line next to the children of nodes that are dumped with the argument set to a true value.  This is hard to explain, so I've posted a gist here: https://gist.github.com/mryan/e64e93362a275fa0b41cf6650c5d8e6c

You see the lines help with seeing the alignment of various nodes to see what's on the same level and what's not.  They also make very clear what's part of any given block.  Having said that, I recognise that this is a debatable improvement - which is why I've said all along " *may* be generally useful".  Please feel free to reject this outright and I'll simply put a patch on a blog somewhere for people to use it if they're interested.  Its just something I conjured up to make the QAST dumps a little easier to read.

### DIFF
--- a/src/QAST/Children.nqp
+++ b/src/QAST/Children.nqp
@@ -18,12 +18,16 @@ role QAST::Children {
     }
 
     method dump_node_list(int $indent, @onto, @node_list) {
+        # Guide lines are triggered from here - if this is a Block
+        # node, then we request a guide line next to its children
+        my $guide-requested := nqp::istype(self, QAST::Block) ;
+
         for @node_list {
             if nqp::istype($_, QAST::Node) {
-                nqp::push(@onto, $_.dump($indent));
+                nqp::push(@onto, $_.dump($indent, :guide-line($guide-requested) ));
             }
             else {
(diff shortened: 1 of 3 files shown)


AlexDaniel:
More readable AST output? Sure, sign me in!

Thank you, @mryan, I have now merged this. The PR was getting a bit old so I rebased it and also fixed a conflict:
* https://github.com/Raku/nqp/commit/185d9a7d2eda2caaed143f2fe3ee5317224605f6
* https://github.com/Raku/nqp/commit/88801ac29f2ed30dd8f3d9044461100a3a011ab5
* https://github.com/Raku/nqp/commit/4eca4bb0f40d728a0e53ecf5f0ff217d300ece8d

If anybody has any problems with this change, please let me know.

PR #455, opened 2018-05-31
Rakudo Issue 1886: Remove erroneous decont op in chain compilation

jstuder-gh:
Removing entirely superfluous decont operation included when compiling
MAST for the chain op. Not only was this decont not doing any good, but
in some scenarios it would result in error (see
[Github Issue 1886](https://github.com/rakudo/rakudo/issues/1886))
although it's unclear to me why it's sometimes a useless addition and
sometimes an explicit error.

### DIFF
--- a/src/vm/moar/QAST/QASTOperationsMAST.nqp
+++ b/src/vm/moar/QAST/QASTOperationsMAST.nqp
@@ -639,7 +639,7 @@ my $chain_gen := sub ($qastcomp, $op) {
 
     # Check if callee sub in name, if not first child is callee, not arg
     my $arg_idx;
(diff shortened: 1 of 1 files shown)


timo:
maybe it caused problems because the decont_reg was being alloced before the loop, but freed every time inside the loop? it could be it was overwriting registers that have been used for other stuff in the mean time perhaps?

jstuder-gh:
@timo, that definitely could the reason for the failure. I figure it's probably safest just to remove the decont altogether. git blame shows that I added it in commit 92d1d56831. It's definitely not necessary for the chaining operation.

PR #462, opened 2018-06-08
[Impl] !~~by adding NOT_ACCEPTS method

### LINKED ISSUE #300: NQP needs "!~~" operator
Originally opened as https://rt.perl.org/Ticket/Display.html?id=128711 by Tom Browder

> NQP currently requires this:
> 
>   unless $var ~~ $regex { ... }
> 
> but it would be convenient to use:
> 
>   if $var !~~ $regex { ... }
>

W4anD0eR96:
In fact, I want this PR, like its name, will be NOT_ACCEPTS.

How we achieve this intention at Rakudo is by introduce a metaop logic and negate the result of EXPR.

Right know if we define `!~~` as infix, then we can add ONLY ONE QAST::Node at pos, thus I add a NOT_ACCEPTS method to implement `!~~`

This will close #300 though

### DIFF
--- a/src/NQP/Actions.nqp
+++ b/src/NQP/Actions.nqp
@@ -1674,6 +1674,10 @@ class NQP::Actions is HLL::Actions {
         make QAST::Op.new( :op<callmethod>, :name<ACCEPTS>, :node($/) );
     }
 
+    method infix:sym<!~~>($/) {
+        make QAST::Op.new( :op<callmethod>, :name<NOT_ACCEPTS>, :node($/) );
+    }
+
     # Takes a multi-part name that we know is in a package and generates
     # QAST to look it up using NQP package semantics.
     sub lexical_package_lookup(@name, $/) {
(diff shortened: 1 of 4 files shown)


W4anD0eR96:
But well, the whole metaop system is somehow too heavy, it's ok too add ONE extra method for this time - If we don't attend to add many non-native ops just negate existing ops.

tbrowder:
Looks good to me, but I’m a real noob in this code.

As an aside: I think that NOT_ACCEPTS could now be NOT-ACCEPTS since NQP can use kebab-case identifiers.

ugexe:
I must be missing something: why this is needed. If we don’t have special negated paths for any other comparison operator? Seems odd to add this without much other commentary ( that I know of anyway ) days before a release.

W4anD0eR96:
> if we don’t have special negated paths for any other comparison operator?

In fact, yes. `gt` `lt` `==` `!=` etc. all have native `nqp::op`, not negate the other op.

W4anD0eR96:
That's why I once think of introduce meta-op system that make `!` as negate meta-op. but it might be somehow too heavy

PR #465, opened 2018-06-08
Modify signals available to getsignals on WIN32

jstuder-gh:
JS:
All the signals we need for Windows are actually already defined in
os.constants.signals

JVM:
Match getsignals JVM with MoarVM/JS backends

### DIFF
--- a/src/vm/js/nqp-runtime/io.js
+++ b/src/vm/js/nqp-runtime/io.js
@@ -553,19 +553,7 @@ op.getsignals = function() {
     return sigCache;
   }
 
-  const osSigs = (function() {
-    if (os.platform() === 'win32') {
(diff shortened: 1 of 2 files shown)


usev6:
@jstuder-gh sorry I've missed your request for a review back in 2018. I don't use Windows, but as far as I can see, adding the signals to ```sigWanted``` on the JVM backend increases consistency and looks sensible. ```nqp::getsignals``` will return negative values for all the not supported signals anyway, right? I can't comment on the changes to the JS backend.

The merge conflict stems from renaming s/perl6/raku/ earlier today (https://github.com/Raku/nqp/pull/630). It shouldn't be hard to resolve -- the changes need probably just copied over to the new location ```src/vm/jvm/runtime/org/raku/nqp/runtime/IOOps.java```.

PR #477, opened 2018-06-29
Add callback of nqp::spawnprocasync to catch exception

### LINKED ISSUE #467: nqp t/nqp/114-pod-panic.t hangs
Hello

Debian runs non-regression tests using the binaries of an installed package. I.e. during these tests, the command `prove  -e nqp t/[chmnpqs]*` in run in freshly checked out nqp repository.

`prove` command hangs on `nqp t/nqp/114-pod-panic.t`. This script launches a nqp subprocess using `./nqp-m` . Since this test is run in a source directory without building, the `nqp-m` file is not present. Unfortunately, the test hangs in this setup.

I would expect this script to exit on error when `./nqp-,` is missing. 

See [Debian bug 900230](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=900230) for more context.

All the best

W4anD0eR96:
This will fix #467 , which now throw an exception with

  no such file or directory
     at gen/moar/stage2/NQPCORE.setting:1226  (NQPCORE.setting.moarvm:)

instead of hanging.

### DIFF
--- a/src/core/testing.nqp
+++ b/src/core/testing.nqp
@@ -114,6 +114,12 @@ sub run-command($command, :$stdout, :$stderr) {
             ++$read-all2;
         }
     };
+    $config<error> := -> $err {
+      my $ex := nqp::newexception();
+      nqp::setmessage($ex, $err);
+      nqp::setpayload($ex, nqp::null());
+      nqp::throw($ex)
+    }
 
     # define the task
     my $task := nqp::spawnprocasync($queue, $command, nqp::cwd(),


W4anD0eR96:
Not really sure whether I did it right, so ask you for review. Thanks!

jnthn:
Yes, due to the fact that the thread that wants the results just loops waiting for events on the `Channel` and processing them, this will throw the exception to the caller of `run-command`, which is the desired semantics.

PR #478, opened 2018-06-29
More robust test

W4anD0eR96:


W4anD0eR96:
@tbrowder I wonder if this is your intention. If we have `./nqp`, then `./nqp` is the right command, isn't it?

W4anD0eR96:
or we can make sure that there is ONLY `./nqp`

ugexe:
The original code seems more correct -- consider if both jvm and moarvm backends are installed.

W4anD0eR96:
@ugexe see #467, there is a situation that only `./nqp` exist.

Could it be
```
my $cmd;
if ((nqp::stat('./nqp-m', nqp::const::STAT_EXISTS) == 0)
    && (nqp::stat('./nqp-j', nqp::const::STAT_EXISTS) == 0)) {
    $cmd := './nqp';
} else {
    $cmd := nqp::getcomp('nqp').backend.name eq 'jvm' ?? './nqp-j' !! './nqp-m';
}
```
or just, at co #477 , throw if `./nqp*` doesn't exist.

ugexe:
Your new code seems reasonable to me.

Maybe https://github.com/perl6/nqp/blob/master/docs/ops.markdown#execname could provide a more robust solution/general solution in the future.

W4anD0eR96:
You're right. It seems what should be improved is `tools/build/gen-xxx-runner.pl`

W4anD0eR96:
close for #479

W4anD0eR96:
MoarVM backend could easily introduce `execname` but JVM backend introduces it (in rakudo) somewhat weird, thus reopen this PR and temporarily to avoid user-block case in #467

PR #479, opened 2018-06-29
Pass --execname to nqp thus nqp::execname() works

W4anD0eR96:


W4anD0eR96:
I'm only test on Mac OS X with MoarVM backend. So make it a pull request to prevent any mistake.

W4anD0eR96:
still I don't know how to pass `execname` variable in JS backend @pmurias

ugexe:
It would be wise to build a rakudo against this and run the roast on it.

W4anD0eR96:
Tested with rakudo/t/* and several roast(I want a `./fudgeandtestall` though), all looks ok with MoarVM backend. It seems JVM backend of Perl 6 impls `execname` by a more different way..

W4anD0eR96:
Only enable `nqp::execname()` on MoarVM backend, I have no idea how to impl it clean in JVM, but I think this is a net win, thus merge this once CI pass. And as mentioned above, it would not break rakudo build and roast test

PR #482, opened 2018-07-04
Display build/commit info on config/make

jstuder-gh:
This will be most helpful using '--gen-nqp=master' when configuring
Rakudo in an environment like Travis or AppVeyor.


tisonkun:
Closed as stale and invalid any more.

PR #483, opened 2018-07-08
Rename getstrfromname to strfromname

MasterDuke17:
This better fits the convention of the other `*from*` ops.

NQP builds ok and passes `make m-test`.


src/vm/jvm/QAST/Compiler.nqp
```diff
@@ -2396,7 +2396,7 @@ QAST::OperationsJAST.map_classlib_core_op('normalizecodes', $TYPE_OPS, 'normaliz
 QAST::OperationsJAST.map_classlib_core_op('codes', $TYPE_OPS, 'codes', [$RT_STR], $RT_INT);
 
 QAST::OperationsJAST.map_classlib_core_op('codepointfromname', $TYPE_OPS, 'codepointfromname', [$RT_STR], $RT_INT);
-QAST::OperationsJAST.map_classlib_core_op('getstrfromname', $TYPE_OPS, 'getstrfromname', [$RT_STR], $RT_STR);
+QAST::OperationsJAST.map_classlib_core_op('strfromname', $TYPE_OPS, 'getstrfromname', [$RT_STR], $RT_STR);
```
W4anD0eR96: You've changed method in Ops.java from `getstrfromname` to `strfromname`, so the next `'getstrfromname'` should be changed, too. I think.

MasterDuke17: Good catch. I've force-pushed a correction.

src/vm/moar/QAST/QASTOperationsMAST.nqp
```diff
@@ -2422,7 +2422,7 @@ QAST::MASTOperations.add_core_moarop_mapping('isge_n', 'ge_n');
 
 QAST::MASTOperations.add_core_moarop_mapping('cmp_s', 'cmp_s');
 QAST::MASTOperations.add_core_moarop_mapping('unicmp_s', 'unicmp_s');
-QAST::MASTOperations.add_core_moarop_mapping('getstrfromname', 'getstrfromname');
+QAST::MASTOperations.add_core_moarop_mapping('strfromname', 'getstrfromname');
```
samcv: Do we want to keep the old mapping and the new mapping? I found it being used in the ecosystem:

https://gist.github.com/Whateverable/3f2e643323a2ec54f7732abd114ebe5d

MasterDuke17: We always say we don't make any promises about NQP (e.g., what ops there are, what they do), so say don't leave the old mapping and I'll just make a PR to that module after the rename happens.

samcv:
Okay, that sounds fine with me.

jnthn:
This doesn't need a new stage0, I don't think.

MasterDuke17:
> This doesn't need a new stage0, I don't think.

It did/does (would get `No registered operation handler for 'strfromname'` when building NQP otherwise), but I'm in the middle of creating a MoarVM PR that renames it there, so when I push that I'll redo this.

Issue #274, opened 2016-01-29
t/nqp/19-file-ops.t failure in pre-2016.01


coke:
This test was failing on OS X for me just prior to the 2016.01 release.

Since we cannot fudge here, commented out the test.

Since this is a new test, should be OK that the behavior isn't conforming to a new test yet.

```
  ### This test was added between 2015.12 and 2016.01, but was failing.
  # since it's something new, commenting out for 2016.01 release.
  #ok(nqp::lstat_time($symlink, $flag) != nqp::lstat_time($file, $flag), 'lstat_time doesn\'t follow symlink');
```

usev6:
Does that test still fail on OS X? It works for me on linux and freebsd for both, MoarVM and JVM.

W4anD0eR96:
@coke How is it going? Can we safely close this issue?

coke:
The test is still commented out since I opened this ticket, so I'm not sure people have been testing it on any platform.

I uncommented it and updated the test count, it's passing for me here on OS X. Someone else can re-open this ticket if needed.

coke:
This test was refudged before the release.

Re-opening ticket.

coke:
See 9a43e6c444

coke:
I unskipped it again, let's see if it stays unskipped for the next release.

jdv:
Skipped again.  What I'm seeing is the nqp test run post nqp build on the release tarball fail on the first run but work subsequent runs...  Looks like so:

t/nqp/019-file-ops.t ................... Failed 1/112 subtests          
	(less 1 skipped subtest: 110 okay)

PR #485, opened 2018-07-19
unfudge test, co #274

IppouTsuukou:
This test pass on my own macOS.

I try to look back the cases that it fails, but surprisingly those did not complain about 019-file-ops.t

strange.

### DIFF
--- a/t/nqp/019-file-ops.t
+++ b/t/nqp/019-file-ops.t
@@ -2,7 +2,7 @@
 
 # Test nqp::op file operations.
 
-plan(109);
+plan(112);
 
 ok( nqp::stat('CREDITS', nqp::const::STAT_EXISTS) == 1, 'nqp::stat exists');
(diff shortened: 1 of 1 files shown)


t/nqp/019-file-ops.t
```diff
@@ -1,4 +1,4 @@
-#! nqp
+#!/usr/bin/env nqp
```
samcv: We may as well remove this line. Since not even all the test files have a shebang line. (Also the files don't have the execute bit either).

TisonShadowsong: done.

I'll appreciate it if you can test this change on a macOS, IIRC the last time @coke uncomment this test, it causes regression. that's why i don't merge this pr yet.

ugexe:
A small nit: "Update $somefile" is not an ideal commit message -- the file name being editing is already part of the commit history and so declaring it has been updated is redundant.

TisonShadowsong:
@ugexe thanks! I will do a "squash and merge" rather than directly merge with every commit.

PR #488, opened 2018-07-22
Implement stubs for getting file descriptors from async sockets

Kaiepi:
I have no idea if this code works or not since I can't run the JVM on my machine, please review


Kaiepi:
Closing this since the way I worked out how to get the native descriptors as soon as the server gets tapped needs to be two different pull requests

PR #499, opened 2018-08-28
add some more nqp examples

tbrowder:


tbrowder:
@lizmat thinks they look ok, merging...

PR #504, opened 2018-09-28
Remove dead code.

patzim:


lizmat:
Hmmm... how can you be sure that that is dead code?

jnthn:
@lizmat Looks like `$prefix` is never used after this point.

lizmat:
Perhaps then `$prefix` should be reduced to `$` in the signature as well ?

zoffixznet:
Doesn't line 13 has a typo and instead of assigning to `$prefix` it should be assigning to `$realpath`?

patzim:
This got stale. I'll have a broader look at this code in general soonish. So close this.

PR #507, opened 2018-10-09
[WIP] [jvm] global clean

TisonKun:
I see the JVM backend is a bit messy and willing to do a global clean, make this pull request to listen if this is supported and run travis tests


TisonKun:
travis fails on a missing `return`, would send a fix with next commit.

jnthn:
This this achieve anything besides making `git blame` less useful and cluttering `final` all over the place on stuff that's effectively final anyway? :/

TisonKun:
For the former, yes, I mix up layout adjustment and code refactor. They are better to be two topic. Sorry.
For the latter, it is a code style thing. Whether place `final` to guard variables are intended to be final, or, use effectively final semantic, save the `final` word and confirm a variable is final by analyze.

PR #510, opened 2018-10-22
[WIP] Fix async socket ops on the JVM

Kaiepi:
I forgot to update the bootstrap files in the original commit and
crucial parts of src/vm/jvm/runtime/org/perl6/nqp/runtime/IOOps.java
were reverted. This allows IO::Socket::Async .listen and .connect to
work on a basic level, but t/05-messages/02-errors.t in Rakudo fails
with a stack overflow.


Kaiepi:
`make spectest` seems to run fine with these changes. Maybe there's a better way to write that one test that fails in particular?

Kaiepi:
Closing since the fix I'm working on now is unrelated to the code from here.

PR #517, opened 2019-01-20
Add uname opcode for MoarVM

ugexe:
Requires https://github.com/MoarVM/MoarVM/pull/1049

### DIFF
--- a/docs/ops.markdown
+++ b/docs/ops.markdown
@@ -301,6 +301,7 @@
   * [const](#const)
   * [cpucores](#cpucores)
   * [getrusage](#getrusage)
+  * [uname](#uname)
   * [debugnoop `jvm`](#debugnoop-jvm)
   * [exit](#exit)
   * [getenvhash](#getenvhash)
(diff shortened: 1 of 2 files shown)


lizmat:
OOC, what are the semantics?

ugexe:
It is modeled after `nqp::getrusage` (of which there is also no docs, so both could share the same template)
https://github.com/MoarVM/MoarVM/blob/918547e6a8f34bf8e44f0ea28bb1573f0f9c6534/src/io/procops.c#L1109
https://github.com/MoarVM/MoarVM/blob/05e2d4222ea47363bea4da8b522b98073bca78b8/src/platform/sys.c#L26-L29

ugexe:
nqp-js implementation should be as simple as:
```
op.uname = function() {
    let sysname = os.GetOSType();
    let release = os.GetOSRelease();
    let version = "";
    let machine = "";
    return [sysname, release, version, machine];
};
```
Not sure how to access the version or machine yet. I'm hoping they will eventually be added to node due to the new availability of this info from libuv. @pmurias

Issue #363, opened 2017-05-29
NQP converts large integers to floats when it shouldn't/doesn't need to


coke:
Originally posted in https://rt.perl.org/Ticket/Display.html?id=131385 by ddgreen @ gmail.com
```
So `nqp-m -e 'say(100000000000000009)'` prints `100000000000000016`.

See https://irclog.perlgeek.de/perl6-dev/2017-05-27#i_14646898 for some discussion with jnthn.

```

zoffixznet:
FWIW, this problem had impacted Perl 6 numerics too ([R#2157](https://github.com/rakudo/rakudo/issues/2157))

In https://github.com/rakudo/rakudo/commit/fec1bd74f9 , I avoided that issue by going under the 2**52 limit under which every integer is representable in double.

Once this issue is fixed, I think we can bump up that limit in Rakudo again.

TisonKun:
Here might be code snip

https://github.com/MoarVM/MoarVM/blob/4d9829c061d35646d045a1d90d27337b3471b89d/src/core/coerce.c#L349-L420

could we auto-detect if we need bigint instead of int in `nqp::radix`?

PR #518, opened 2019-02-04
Make variables and ops default to int

MasterDuke17:
Requires https://github.com/MoarVM/MoarVM/pull/1038.

NQP builds ok and passes `make m-test` (with the edited tests).

Fixes https://github.com/perl6/nqp/issues/363.


MasterDuke17:
@pmurias I can't get the JS backend to build to test this against. Does it work for you?

PR #521, opened 2019-03-02
Add given value to (s)printf bad-directive-type error

MasterDuke17:
NQP builds ok and passes `make m-test`.


lizmat:
Thank you!

PR #522, opened 2019-03-16
Add stack size parameter to nqp::newthread

ugexe:
Adds a stack size parameter to set the initial stack size of a new thread.

Requires https://github.com/MoarVM/MoarVM/pull/1069


usev6:
@ugexe unfortunately merging the s/perl6/raku rename (https://github.com/Raku/nqp/pull/630) caused a conflict for this PR. If I'm not mistaken the change to ```src/vm/jvm/runtime/org/perl6/nqp/runtime/Ops.java``` just needs to go into ```src/vm/jvm/runtime/org/raku/nqp/runtime/Ops.java```.

PR #523, opened 2019-03-16
[jvm, js] Tweak check for illegal range endpoint

usev6:


usev6:
@pmurias Maybe you could look if tweaking the nqp::chars check works for the js backend? As far as I can see it works for the jvm. That way the check for illegal range endpoint could stay.

usev6:
I forgot to mention it: This should help with https://github.com/rakudo/rakudo/issues/2748

usev6:
I've committed something else, that doesn't affect the js backend: 0e5f3d0b26

Issue #528, opened 2019-04-06
bash should not be a required dependency


Kaiepi:
`bash` is not installed by default on *BSD. `sh` should be used instead in `tools/build/install-moar-runner.pl`

patzim:
I'm on this.

PR #529, opened 2019-04-06
Make tools/build/install-moar-runner.pl possible to run on *BSD

Kaiepi:
Related to #528, but doesn't entirely fix it since it still uses bash.


jnthn:
Is this still required now #530 has been merged?

Kaiepi:
No.

PR #530, opened 2019-04-06
Don't use bash

patzim:
Fixes perl6#528. Replaces all uses of bash with sh.


AlexDaniel:
```
        # Note: We parse `ls -l` output to find the symlink target
        #       which is the only POSIX-compliant, albeit somewhat fragile, way.
```
Oh gawd.

Anyway, @patzim++.

PR #532, opened 2019-04-24
Change default profile options

MasterDuke17:
Now `--profile(-compile)=<foo>` will set the filename and
`--profile-kind=<foo>` will set whether it's an instrumented or heap
profile.

If this is merged I'll create a corresponding rakudo PR to correct its help output.

### DIFF
--- a/src/vm/moar/HLL/Backend.nqp
+++ b/src/vm/moar/HLL/Backend.nqp
@@ -59,7 +59,15 @@ class HLL::Backend::MoarVM {
         }
     }
     method run_profiled($what, $filename, $kind) {
-        $kind := 'instrumented' unless $kind;
+        unless $kind {
+            if $filename ~~ / '.html' | '.json' | '.sql' $ / {
+                $kind := 'instrumented';
+	    } elsif $filename ~~ / '.mvmheap' $ / {
+                $kind := 'heap';
+	    } else {
+                $kind := 'instrumented';
+            }
+        }
 
         self.ensure_prof_routines();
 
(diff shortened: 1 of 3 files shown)


timo:
could you put in a little heuristic that selects instrumented when the filename ends in .html, .json, or .sql and heap when it ends in .mvmheap?

jnthn:
Could this break existing tools that invoke the profiler?

MasterDuke17:
> Could this break existing tools that invoke the profiler?

@jnthn it could, depending on how they actually invoke it. If they just do `--profile` then everything would be the same. If they do `--profile=(heap|instrumented)` or `--profile-filename=<foo>` then things won't work as they expect. The only tool I know of that does any profiling is Comma, I assume you guys could adapt it to this change (and no reason it couldn't be synchronized with a Comma release if you desire).

If you or anybody else has strong objections this doesn't have to be merged. It's just that I routinely want to use `--profile-filename=<foo>`, but almost never use `--profile=heap`, so I thought this would be a Huffmanization of the options.

MasterDuke17:
> could you put in a little heuristic that selects instrumented when the filename ends in .html, .json, or .sql and heap when it ends in .mvmheap?

Good idea. Might not get to it today, but I'll add a commit that does that soon.

MasterDuke17:
@timo done (finally).

src/vm/moar/HLL/Backend.nqp
```diff
@@ -58,24 +58,35 @@ class HLL::Backend::MoarVM {
...
+                $kind := 'instrumented';
+            }
+        }
 
         self.ensure_prof_routines();
 
         if $kind eq "heap" {
-            unless nqp::defined($filename) {
+            unless $filename {
                 $filename := 'heap-snapshot-' ~ nqp::time_n();
```
jnthn: Not related to your change, but while we're at it: this may want a `.mvmheap` extension.

PR #536, opened 2019-05-13
[JVM] Don't warn about illegal reflective access

usev6:
... on Java 9 and newer.


usev6:
This avoids warnings when using Rakudo. It's not urgently needed for NQP, since the problematic code isn't excercised by any test.

PR #540, opened 2019-05-18
Parse |, ||, &, && prefixes not in nibble

Altai-man:
Needed for https://github.com/rakudo/rakudo/pull/2907 to work.

### DIFF
--- a/src/QRegex/P6Regex/Grammar.nqp
+++ b/src/QRegex/P6Regex/Grammar.nqp
@@ -132,15 +132,6 @@ grammar QRegex::P6Regex::Grammar is HLL::Grammar {
             for $OLDRX { %*RX{$_.key} := $_.value; }
         }
         <.ws>
-        [
-          <!rxstopper>
-          [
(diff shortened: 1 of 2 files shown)


Altai-man:
Passes spectest.

Altai-man:
It dies due to incorrect test that wants the opposite of what roast wants, let me check it further...

Altai-man:
So we have a test in roast
https://github.com/perl6/roast/blob/master/S05-mass/rx.t#L828
added by Larry at https://github.com/perl6/roast/commit/ee1bd8d63256affa75ffb2753cbdb58f6ad039f6#diff-4149f4a4f6ad1730d0d001f26fd21c0b
that is exactly opposite to what was in nqp, so I set the test to be allowed.

It seems that earlier it "worked" due to incorrect assumption in the test, and the opposite test in roast worked because there was an another bug, that was fixed at https://github.com/rakudo/rakudo/pull/2907/files

PR #543, opened 2019-05-24
support build when moar's prefix ne target prefix

kazcw:
Add a new config option, --with-moar-root=$DIR. This is necessary
to be able to build moar and nqp as separate NixOS packages.

### DIFF
--- a/tools/lib/NQP/Config/NQP.pm
+++ b/tools/lib/NQP/Config/NQP.pm
@@ -170,7 +170,7 @@ sub moar_config {
     return $moar_config if $moar_config->{moar};
     my $sdkroot = $self->cfg('sdkroot');
     my $prefix  = $self->cfg('prefix');
-    my $moar_prefix =
+    my $moar_prefix = $self->opt('with-moar-root') or
       $sdkroot ? File::Spec->catdir( $sdkroot, $prefix ) : $prefix;
     my $moar_exe = $self->opt('with-moar')
       || File::Spec->catfile( $moar_prefix, 'bin', "moar" . $self->cfg('exe') );


kazcw:
This is not sufficient. On unix install-moar-runner.pl still assumes moar and nqp are installed to the same directory.

kazcw:
I think it'll be simpler to apply nix-specific patches than add the necessary knobs.

niner:
But rakudo has supported different prefixes for MoarVM, NQP and rakudo for years via the --with-moar and --with-nqp options respectively?

kazcw:
@niner I'm trying to create separate packages for moar and nqp (and later rakudo), so I can't make use of any rakudo options to get nqp to build. nqp has its own with-moar option, but that only sets the path to the moar binary, not the support files that go with it, which is what this PR was in reference to. What I ended up doing is adding a patch like this to enable configuring the location of the whole moar installation, and another patch to set `$relocatable = 0` in `install-moar-runner.pl`, and I now have `nqp` successfully building against a `moar` in another directory.

niner:
I've filed issues for both nqp and rakudo. This is clearly a regression that needs fixing before the next release.

vrurg:
@kazcw From you mentioning `install-moar-runner.pl` I conclude that you're not working on `master`. The script is gone already.

kazcw:
@vrurg I'm confused. I'm running de6e46c, which is currently the head of the branch called "master" in this repo. There's a file tools/build/install-moar-runner.pl: https://github.com/perl6/nqp/blob/master/tools/build/install-moar-runner.pl. It seems to be actively used, because by modifying it I can change the `nqp` and `nqp-m` scripts that get generated. Am I on the wrong `master` somehow?

vrurg:
@kazcw That's my _oops_, I apologize for the confusion. While actively switching between `nqp` and `rakudo` builds I mixed them up and was speaking about `rakudo` at that moment.

kazcw:
Oh that makes sense. I thought it seemed like we were talking about two different repos!

PR #553, opened 2019-06-07
Implement freemem and totalmem ops

Kaiepi:


Kaiepi:
See https://github.com/rakudo/rakudo/pull/2959

PR #554, opened 2019-06-10
Use --release 9 instead of -source 1.9 in the JVM's Makefile

Kaiepi:
This shuts up warnings about the target being the wrong version of the
JDK and warnings about -bootclasspath being needed.


usev6:
Thanks!

PR #557, opened 2019-06-13
Change associativity of exponentiation to 'right'

taboege:
Just stumbled on this:

https://github.com/perl6/nqp/blob/7cccdb13e28b2aea7321bdef02408f842183d477/src/NQP/Grammar.nqp#L747

Exponentiation should be right-associative (like it is in Perl 6), no? The test suite passed before and passes after this PR, so I am not sure if this will have fall-out downstream. The second commit adds tests for the `**` operator which hasn't been exercised at all.

Maybe this is an instance of "if it's not (too) broken, don't fix it"?

### DIFF
--- a/src/NQP/Grammar.nqp
+++ b/src/NQP/Grammar.nqp
@@ -744,7 +744,7 @@ grammar NQP::Grammar is HLL::Grammar {
 
     my %methodop       := nqp::hash('prec', 'y=', 'assoc', 'unary');
     my %autoincrement  := nqp::hash('prec', 'x=', 'assoc', 'unary');
-    my %exponentiation := nqp::hash('prec', 'w=', 'assoc', 'left');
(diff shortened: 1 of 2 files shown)


tbrowder:
That looks good to me--good catch! You should be able to merge it yourself.

Thanks!

taboege:
Ok, just to be extra, extra sure I built rakudo and ran its test suite and the spectest with these patches applied to the backing nqp. Really doesn't seem to break anything.

taboege:
@tbrowder : No, apparently I have no write permissions to nqp, so I can't merge it myself.

b2gills:
I'm sure that no one has ever tried using two or more exponentiation operators in NQP.

taboege:
And finally we know why :-)

tbrowder:
Thanks!

PR #558, opened 2019-06-21
[IP6NS Grant] Make socket family handling portable

Kaiepi:
MoarVM requires that the socket family be a separate parameter for
nqp::bind_sk and nqp::connect_sk from the port; update the JVM
accordingly and use the same platform-independent constants here as
in MoarVM.

Related to https://github.com/rakudo/rakudo/issues/3007

This still needs testing.


Kaiepi:
Now passes `make test`, perl6-j also passes `make test`, `make spectest`

PR #562, opened 2019-07-05
Add a note about the JVM and ulimit to the README

Kaiepi:


usev6:
The added note looks good to me. I wonder if it would make sense to insert it a bit further down, though:
a) Since the JVM backend is classified as 'experimental', I'd expect specific notes for this backend to come after notes for MoarVM.
b) With regard to its contents the note would fit under ```troubleshooting```.

coke:
@Kaiepi - can you address @usev6 's comments so we can merge this?

usev6:
Thanks a lot!

PR #564, opened 2019-07-11
Implement NativeCall wide string support

Kaiepi:
See the Rakudo pullreq for info on what this does.


usev6:
@Kaiepi and another of your open PR that has conflicts after I merged the s/perl6/raku/ rename (https://github.com/Raku/nqp/pull/630). Again, it's hopefully easy to solve the conflicts, since the changes just need to be copied to the new locations of the changed files (```src/vm/jvm/runtime/org/nqp/...```.

I don't feel qualified to comment on the changes you made.

lizmat:
It feels like all of Kaiepi's work will be lost unless someone reimagines this work.

@patrickbkr @timo @niner

PR #569, opened 2019-08-05
Fix for a rare case of empty mast_frames hash

vrurg:
Caused by a bug reported in rakudo/rakudo#3096. This fix is somewhat of
a guess because I'm not familiar with QAST->MAST compilation. But the
logic I put behind it: if no frames found in mast_frames then just do
nothing.

### DIFF
--- a/src/vm/moar/QAST/QASTOperationsMAST.nqp
+++ b/src/vm/moar/QAST/QASTOperationsMAST.nqp
@@ -603,8 +603,10 @@ QAST::MASTOperations.add_core_op('list_b', -> $qastcomp, $op {
                 unless nqp::istype($_, QAST::Block);
             my $cuid  := $_.cuid();
             my $frame := $qastcomp.mast_frames{$cuid};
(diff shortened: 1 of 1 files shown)


vrurg:
> I do not think this is the right approach. It would simply skip adding those static code refs to the serialization context. It may hide some immediate errors, but cause more obscure ones later on.

That's why I requested for a review. I felt like it's not something correct even though it made the `EVAL` work. Still, I wonder if there should be a rather simple fix. Perhaps, in _ForeignCode.pm6_. Considering line https://github.com/rakudo/rakudo/blob/master/src/core/ForeignCode.pm6#L54 and https://github.com/rakudo/rakudo/blob/master/src/core/ForeignCode.pm6#L74 I guess that the empty hash somehow makes it all they way down to the `list_b` code. I guess a frame must be added to that hash then. But which one? `$compiled`?

> Note that compile time eval is only somewhat implemented. There are still missing bits and pieces and hacks in place of real implementations, so a quick and easy fix may just not do in this case.

Noted. I'm closing this PR as wrongful.

PR #570, opened 2019-08-05
set -municode in LD_M_BUILD_RUNNER to fix the MinGW build

cygx:


patzim:
Looking good. Thanks cygx++!

PR #572, opened 2019-08-12
Enable dynamic QAST Tree building

KennethTrecy:
By using function, the developer can plug different types of QAST node in the AST. For example, upon invoking the function, it may return `QAST::Op.new(:op<callmethod>)` or `QAST::Var.new(:scope<attribute>)`, depending on the terms that was passed.


AlexDaniel:
Hello! I invited you to the perl6 org in case you have more fixes :) As for this one, I think we'd need to wait for somebody else to review.

PR #584, opened 2019-10-14
Produce better error message when callee is not found

vrurg:
Currently if callee is not found we end up with LTA message _'Cannot invoke this object (REPR: Null; VMNull)'_. Instead, if caused by a symbol missing in callers lexical scope, it would now result in something like _'Cannot find callee &prefix:<++> in the lexical scope'_

rakudo/rakudo#2897

### DIFF
--- a/src/vm/moar/QAST/QASTOperationsMAST.nqp
+++ b/src/vm/moar/QAST/QASTOperationsMAST.nqp
@@ -1487,11 +1487,31 @@ my $call_gen := sub ($qastcomp, $op) {
     my $return_type;
     my @args := $op.list;
     if $op.name {
-        $callee := $qastcomp.as_mast(QAST::Op.new(
-            :op('decont'),
-            $op.op eq 'callstatic'
-                ?? QAST::VM.new( :moarop('getlexstatic_o'), QAST::SVal.new( :value($op.name) ) )
-                !! QAST::Var.new( :name($op.name), :scope('lexical') )));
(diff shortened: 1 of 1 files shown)


jnthn:
I'm not sure we can afford to do it this way, really. The bytecode size of a sub call is probably one of the most important things to keep down so far as code size goes, because they're simply *everywhere* - even operators are sub calls. (They also play into the inline budget, so this could throw off a bunch of optimizations too.) Further, this op is emitted by the optimzier when it knows the lookup should really be successful; something has gone very wrong (usually precomp) if the lookup fails. Perhaps we should experiment with making the VM-level op throw if it can't perform the lookup instead.

vrurg:
Ok, you have confirmed one my suspicion and added one more on top of it.

I was looking into a possible way to have it in MoarVM. But not sure how to do it properly. The point is that it's the only moment in `call` when symbol name is known and thus informative error message could be produced. So:

- `getlextstatic_o` must not throw. Or it needs additional parameter-flag, but this would break other code. Or a new op with stricter semantic is needed.
- Unfortunately, I don't know yet what `QAST::Var` is mapped into. Not a big deal, but somewhat problematic for me now.

PS. I'd keep this PR for a while just for the discussion itself.

vrurg:
Ah, and – yes, it is still about the precompilation. Somehow core setting is getting lost and only unit's scope is captured.

Issue #589, opened 2019-12-09
[WIP] JVM build needs a couple of fixes.


vrurg:
- Dependency on runtime .java files.
- configure_jars shall be moved from `NQP::Configure` in `nqp-configure` submodule to `NQP::Configure::NQP`.

This task just an anchor for WIP.

pmurias:
The problem is that nqp-j uses  gen/jvm/sshare/runtime/nqp-runtime.jar which is copied from nqp-runtime.jar but does not depend on it. So when a java file is changed and recompiled nqp-j still uses the old runtime from before recompile.

vrurg:
Closing via #590 and #591

PR #590, opened 2019-12-09
Make NQP::Configure::NQP take care of jars

vrurg:
Configure.pl must contain only what's absolutely necessary.


vrurg:
Related to #589

PR #597, opened 2020-01-30
[IP6NS Grant] Implement support for bind before connect with sync sockets

Kaiepi:
This is needed for upcoming support for bind before connect and UDP sockets with `IO::Socket::INET`.

Passes `make test` and `make spectest`.

See https://github.com/MoarVM/MoarVM/pull/1231

### DIFF
--- a/src/vm/jvm/runtime/org/perl6/nqp/io/AsyncServerSocketHandle.java
+++ b/src/vm/jvm/runtime/org/perl6/nqp/io/AsyncServerSocketHandle.java
@@ -17,25 +17,32 @@ import org.perl6.nqp.sixmodel.reprs.AsyncTaskInstance;
 import org.perl6.nqp.sixmodel.reprs.ConcBlockingQueueInstance;
 import org.perl6.nqp.sixmodel.reprs.IOHandleInstance;
 
-public class AsyncServerSocketHandle implements IIOBindable, IIOCancelable {
+public class AsyncServerSocketHandle implements IIOBindable, IIOListenable, IIOCancelable {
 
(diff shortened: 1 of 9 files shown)


src/vm/jvm/runtime/org/perl6/nqp/io/AsyncServerSocketHandle.java
```diff
@@ -17,25 +17,32 @@
...
+            throw ExceptionHandling.dieInternal(tc, "Failed to resolve host name");
+    }
+
+    @Override
+    public void listen(ThreadContext tc, int backlog) {
+        if (addr == null)
+            throw ExceptionHandling.dieInternal(tc, "Socket must be bound before it can be listened on");
         try {
-            InetSocketAddress addr = new InetSocketAddress(host, port);
             listenChan.bind(addr, backlog);
```
jnthn: Does this actually move the point of binding on JVM, or is that simply not possible?

Kaiepi: Binding in Java does both binds and listens on the socket. There's no way to do the binding in the bind op since it doesn't take a backlog argument, so I decided to make it resolve the address on bind, then bind on listen.

usev6:
@Kaiepi and another of your open PR that has conflicts after I merged the s/perl6/raku/ rename (https://github.com/Raku/nqp/pull/630). Again, it's hopefully easy to solve the conflicts, since the changes just need to be copied to the new locations of the changed files (```src/vm/jvm/runtime/org/nqp/...```.

I don't feel qualified to comment on the changes you made.

PR #598, opened 2020-02-11
Don't fail with `--git-reference` Configure.pl parameter

patrickbkr:
This makes it backwards compatible. The option has been a no-op for quite
some time, so we don't need to actually implement it.

@vrurg: I'm not sure I picked the right spot for that warning message. Do tell if it's better placed elsewhere.


patrickbkr:
This has been solved differently.

PR #601, opened 2020-03-01
Add nextdispatcherfor/takenextdispatcher ops

vrurg:
Provide support for Raku chained dispatchers. Their purpose is to pass
information to the downstream dispatcher about what upstream dispatcher
must take over the control next when downstream exhausts.

For example, when one wraps a candidate in a multi, the `WrapDispatcher`
must know about the instance of `MultiDispatcher` to switch back to it
when all wrappers are done and the candidate calls one of
`{next|call}{same|with}`.

Technically, both ops are clones for `setdispatcherfor`/`takedispatcher`
except that they operate with `next_dispatcher` and
`next_dispatcher_for` members of `ThreadContext`; and their purpose
differs too.

Requires MoarVM/MoarVM#1252.


vrurg:
@pmurias May I ask you for JVM implementations of the ops?

pmurias:
> @pmurias @MasterDuke17 May I ask you for JVM implementations of the ops? I could try cloning the sources of `setdispatcherfor`/`takedispatcher`, but not sure if I shall touch the language I don't know.

I could implement this for JVM and JS

vrurg:
@pmurias Thanks! Though JVM is done already, unless you think I did it a wrong way.

pmurias:
At a quick glance the JVM implementation seems reasonable, I'll implement it for the *JS* backend.

PR #605, opened 2020-03-14
Introduce VMNull on JVM backend

usev6:
Again and again we have to add special cases in Rakudo for the JVM backend because ```nqp::null()``` creates a plain Java ```null``` whereas MoarVM has a concept of a VMNull.

The goal of this branch is to bring the JVM backend in line with MoarVM and let ```nqp::null()``` create a SixModelObject VMNull. In order to do that I changed ```isnull``` so that it checks for VMNull, too. Also, I've changed most of the ```null``` checks for SixModelObjects to use the pimped ```isnull``` instead.

I might have missed some ```null``` checks. Also, it could very well be the case that some of the changed checks shouldn't look for VMNull. But this branch passes all NQP tests (except some fudged tests) -- like current master does. Also, with some minor tweaks to JVM specific code in Rakudo, this branch seems to pass nearly all of the spectests working on NQP master.

### DIFF
--- a/src/vm/jvm/QAST/Compiler.nqp
+++ b/src/vm/jvm/QAST/Compiler.nqp
@@ -1073,7 +1073,11 @@ QAST::OperationsJAST.add_core_op('ifnull', -> $qastcomp, $op {
     my $lbl := JAST::Label.new( :name($qastcomp.unique('ifnull_')) );
     $*STACK.obtain($il, $expr);
     $il.append($DUP);
-    $il.append(JAST::Instruction.new( :op('ifnonnull'), $lbl ));
+    $il.append(JAST::Instruction.new( :op('invokestatic'),
+        $TYPE_OPS, 'isnull', 'Long', $TYPE_SMO));
+    $il.append($IVAL_ZERO);
+    $il.append($LCMP);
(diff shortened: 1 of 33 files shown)


lizmat:
I can't judge the validity of the code, but I very much welcome the effort!

usev6:
With the latest commit (3a4debadc7) I see only those spectest failures that already existed on NQP master.

Of course, it would be nice to get a review (if only a short evaluation of the idea to add a SixModelObject VMNull the way I did). On the other hand other things are probably more important and maybe it doesn't make sense to invest scare resources like @jnthn's time in this topic.

I tend to just push the merge button later this week if no-one objects or comments. There is always the option to revert ...

pmurias:
Looks sane to me. Maybe we could use a null singleton or a helper function instead of == 1 for the null check but that can be done after the merge.

usev6:
Thanks for looking. I'm going for the merge now.

@pmurias I used the existing Ops.isnull method (that returns a long) to avoid bootstrap issues. (I remember a warning from peschwa about that.) I'd be glad to improve on this, but I didn't see a better way. Could you maybe sketch out how a null singleton could look like / be used?

PR #606, opened 2020-03-19
[JVM] Always provide file and line in backtrace

usev6:
This helps with a couple of spectests that die in core.c/Backtrace.pm6, in
https://github.com/rakudo/rakudo/blob/9429728cbd/src/core.c/Backtrace.pm6#L107

The code tries to mimic the behaviour of MoarVM, esp.
https://github.com/MoarVM/MoarVM/blob/3c3ad0678a/src/core/exceptions.c#L555-L572


usev6:
Spectest looked better than without this change, so I'll merge it.

PR #608, opened 2020-03-30
Use NQP prefix for finding profiler template

### LINKED ISSUE #567: --profile uses wrong prefix to find template.html
This line uses the MoarVM prefix to find template.html, whereas the file is in fact under the NQP prefix:

chloekek:
Fixes #567.

This fixes the problem but I am not sure what would be the best way to generate the file with the nqpconfig subroutine definition in it.

### DIFF
--- a/tools/lib/NQP/Config/NQP.pm
+++ b/tools/lib/NQP/Config/NQP.pm
@@ -93,6 +93,14 @@ sub configure_misc {
 
     $config->{moar_stage0} = $self->nfp( "src/vm/moar/stage0", no_quote => 1 );
     $config->{jvm_stage0}  = $self->nfp( "src/vm/jvm/stage0",  no_quote => 1 );
+
+    # TODO: Use the template infrastructure from nqp-configure?
+    open my $configfile, '>', 'src/core/Config.nqp';
+    print $configfile "# This file is generated automatically by $0\n";
+    print $configfile "sub nqpconfig() {\n";
+    print $configfile "    nqp::hash('prefix', '$config->{prefix}');\n";
+    print $configfile "}\n";
+    close $configfile;
 }
 
 sub configure_moar_backend {
--- a/src/vm/moar/HLL/Backend.nqp
+++ b/src/vm/moar/HLL/Backend.nqp
@@ -13,7 +13,7 @@ my sub literal_subst(str $source, str $pattern, $replacement) {
 
 class HLL::Backend::MoarVM {
     our %moar_config := nqp::backendconfig();
-    
+
     my sub read_ui32($fh, $buf?) {
         unless $buf { $buf := nqp::create($NQPBuf) }
         nqp::readfh($fh, $buf, 4);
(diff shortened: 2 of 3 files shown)


tools/lib/NQP/Config/NQP.pm
```diff
@@ -93,6 +93,14 @@ sub configure_misc {
 
     $config->{moar_stage0} = $self->nfp( "src/vm/moar/stage0", no_quote => 1 );
     $config->{jvm_stage0}  = $self->nfp( "src/vm/jvm/stage0",  no_quote => 1 );
+
+    # TODO: Use the template infrastructure from nqp-configure?
+    open my $configfile, '>', 'src/core/Config.nqp';
```
vrurg: We must not autogenerate anything in src. It actually belongs to `gen/`.

vrurg:
It seems to me that the correct code in `HLL/Backend.nqp` should be:

```raku
my $temppath := nqp::getcomp("nqp").config()<libdir> ~ '/profiler/template.html';
```

chloekek:
@vrurg I gave that a try. It fails with: _X::AdHoc: Cannot call method 'config' on a null object_. However, method config is available on HLL::Compiler, but _it_ calls HLL::Backend::MoarVM.

vrurg:
Ok, I see. I didn't know that rakudo uses profiling code from nqp. `nqp::getcomp("nqp")` makes sense under nqp runtime only. For rakudo the line would look like:

```raku
nqp::getcomp("Raku").config<nqp-home> ~ 'lib/profiler/template.html'
```

So, the problem would be to know which compiler to use. @lizmat do you have an idea?

lizmat:
Try both, and see which one yields something to work with?

vrurg:
@lizmat the first thought. But what if somebody writes another nqp-based compiler?

Actually, as I look into it, the place where template.html gets installed is incorrect because the file is moar-backend specific and as such should be installed under some kind of backend subdir.

vrurg:
So far, it looks like the proper solution would be to add compiler parameter to `run_profiled`, `dump_profile_data`, and `dump_instrumented_profile_data` methods. Then instead of `nqp::getcomp(...).config` `$compiler.config` can be used.

Then, there is minor discrepancy to be taken care of somehow. NQP config has `nqp_home` key. But in Rakudo it is named `nqp-home`. I think this should be solved by renaming `nqp_home` into `nqp-home` or providing both variants for nqp compiler.

lizmat:
If `nqp_home` is something that can be set by environment variables, then I would choose to to rename the `nqp-home` to `nqp_home` instead of vice-versa: you cannot use hyphens in environment variables afaik.

vrurg:
No, it is purely our implementation detail thing. It is only affected by _Configure.pl_ command line. If there is 

Changing it to `nqp-home` would also go in line with `source-digest` key in the compiler config.

chloekek:
> So far, it looks like the proper solution would be to add compiler parameter to run_profiled, dump_profile_data, and dump_instrumented_profile_data methods. Then instead of nqp::getcomp(...).config $compiler.config can be used.

This seems overly complicated. When configuring NQP, we already know where it is going to be installed into, so we should be able to just get the path in there.

What about making hll-config, which is the function generated by gen-version.pl, part of the core setting? Then we can just call that and extract nqp-home or even libdir.

chloekek:
Nevermind, this has to be done because nqp-home can be overridden with the NQP_HOME environment variable when invoking Rakudo. I’ll see if passing the compiler config around works.

lizmat:
And if the environment variable is called NQP_HOME, I think it should be called "nqp_home" internally as well.

vrurg:
@lizmat I disagree. By this logic we should use uppercase too. As well as `rakudo-home` key.

Snake case would make it inconsistent with other keys in the config. It's even more confusing, than the difference from the environment. Whereas if one already knows what's the rule of thomb he'd expect the key to be kebab-lower-cased.

PR #612, opened 2020-04-05
Write large chunks of bytecode to a file on the fly instead of doing it all in memory first

MasterDuke17:
A Rakudo compile doesn't take any longer, but heaptrack reports 4gb fewer bytes allocated (just for compiling CORE.c). Also use the new ability of `nqp::encode` to append to a pre-existing buffer when building the string heap.

Requires https://github.com/MoarVM/MoarVM/pull/1267


src/vm/moar/QAST/QASTCompilerMAST.nqp
```diff
@@ -507,7 +507,7 @@ my class MASTCompilerInstance {
                 if $got == $MVM_reg_int64 {
                     %core_op_generators{'coerce_is'}($res_reg, $reg);
                 }
-                elsif $got == $MVM_reg_int32 {
+                elsif $got == $MVM_reg_int32 || $got == $MVM_reg_int16 || $got == $MVM_reg_int8 || $got == $MVM_reg_uint32 || $got == $MVM_reg_uint16 || $got == $MVM_reg_uint8 {
```
niner, line 510: This is probably correct but a bit confusing. Took me a second read to see that since we're coercing to int64 first, the coerce_is is correct (as opposed to a coerce_us like for MVM_reg_uint64). And it's correct, since we're handling only types that are smaller than int64, so it can't overflow.

src/vm/moar/QAST/QASTCompilerMAST.nqp
```diff
@@ -2614,6 +2631,67 @@ class MoarVM::BytecodeWriter {
         self.write_annotations;
         self.align;
     }
+    method assemble_to_file($filename) {
+        my $file := nqp::open($filename, 'w');
+        nqp::closefh($file);
+        $file := nqp::open($filename, '+');
+
+        # write the header, but we'll have to go back and correct some of the sizes
```
niner, line 2639: Why don't you skip the header and write it last? It's always 92 bytes + 4 bytes padding long, so skipping is easy. That would get rid of the scary looking update code with those magic numbers further down.

src/vm/moar/QAST/QASTCompilerMAST.nqp
```diff
@@ -2614,6 +2631,67 @@ class MoarVM::BytecodeWriter {
...
+        # read the old bytecode offset, bytecode size, and annotations offset
+        my $tmp := MAST::Bytecode.new;
+        nqp::readfh($file, $tmp, 12);
+        my int $prev_bytecode_offset := $tmp.read_uint32_at(0);
+
+        # write the correct bytecode size and new annotations offset
+        $tmp.write_uint32_at($bytecode_size, 4);
+        $tmp.write_uint32_at($prev_bytecode_offset + align_section($bytecode_size), 8);
+        nqp::seekfh($file, 13*4 + 8, 0);
+        #nqp::seekfh($file, 11, 1);
```
niner, line 2684: commented out code

Issue #616, opened 2020-04-15
Remove unused jvm opcodes


coke:
Based on the list of undocumented JVM opcodes (and then grepping through the nqp & rakudo source, as well as the greppable6 output), I found the following list of seemingly unused opcodes:

```
closefh_i
linesasync
nfatostatelist
openasync
printfh
readallfh
readcharsfh
readlinechompfh
readlinefh
setencoding
setinputlinesep
shell
slurpasync
spawn
spurtasync
syncpipe
```
I'll put together a branch to remove all these opcodes. @jnthn did mention on IRC that several of these are definitely leftovers.

PR #617, opened 2020-04-16
JVM opcode cleanup

coke:
With these removals, still able to build/test jvm backend on both NQP & Rakudo


coke:
For #616

coke:
When merging, these can be squashed, just wanted to be able to test them out as I went.

PR #622, opened 2020-04-24
Fix building JVM on Windows

patrickbkr:


vrurg:
The exec line in nqp-j looks the same to me, it's only the setenvs which differ. I would rather wrap them in `@if(platform==w indows ...)@` and `platform!=windows`.

patrickbkr:
@vrurg I adapted the templates accordingly. Thanks!

PR #623, opened 2020-04-24
Enable running precompiled scripts

patrickbkr:
This change is part of the GSoC of @pamplemoussecache but is useful on its own.
With this change it is possible to run a precompiled script:
```
raku --target=mbc --output=script.mbc script.raku
raku -b script.mbc
```
It is possible there limits to this functionality that are currently unexplored, but I think we should get this change merged anyways.

### DIFF
--- a/src/HLL/Compiler.nqp
+++ b/src/HLL/Compiler.nqp
@@ -411,8 +411,8 @@ class HLL::Compiler does HLL::Backend::Default {
         $!user_progname := join(',', @files);
         my @codes;
         for @files -> $filename {
-        my $err := 0;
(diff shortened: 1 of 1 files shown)


lizmat:
Ideally, a script should be pre-compiled if it has a `use precompilation` in it.  I guess this patch brings that closer to a reality?

patrickbkr:
@lizmat I think so. Though I guess there is still some PrecompRepo work necessary to make it happen.

patrickbkr:
With this PR it is not possible to run compiled scripts that use modules. More work is necessary.
So *no* merge for now.

PR #638, opened 2020-05-23
Fix sprintf() with *-specified negative width argument

softmoth:
It's already documented like this in `p6doc independent-routines`, and matches C behavior.

### DIFF
--- a/src/HLL/sprintf.nqp
+++ b/src/HLL/sprintf.nqp
@@ -148,7 +148,7 @@ my module sprintf {
 
         sub padding_char($st) {
             my $padding_char := ' ';
-            if (!$st<precision> && !has_flag($st, 'minus'))
+            if !$st<precision> && !is_minus($st)
             || $st<sym> ~~ /<[eEfFgG]>/ {
                 $padding_char := '0' if $_<zero> for $st<flags>;
             }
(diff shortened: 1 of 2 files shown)


tbrowder:
That looks good to me. Does that enhancement require anything special to sprintf in Raku to be used there? (I'm trying to learn so I can help improve sprintf, too.)

lizmat:
Nothing needs to be done to make Rakudo use this `sprintf`, as Rakudo uses the nqp `sprintf` currently.

I've looked at the Vector flag and could not see a use for that in Raku

lizmat:
Thanks!

softmoth:
Thanks, Tom & Elizabeth. As for the vector flag, my non-expert opinion is it would be precisely as useful in Raku as it is in Perl. I don't think there are significant Raku features that obsolete the vector flag. I also am not sure that the flag is all that useful in Perl, where explicitly showing the work with `map` and `join` is probably fine. So I'd personally be OK letting Perl have a feature that Raku lacks, here.

Another drive-by comment about nqp::sprintf is that the behavior of the `0` flag is wrong (in my opinion) with %0s, %0c, and any other non-numeric formats. In C the `0` flag is ignored (gcc prints a warning at compile time). I think it's not an improvement that Raku pads with 0, and I'd recommend changing that to match C's behavior.

tbrowder:
I hope to get a makeover for our sprintf doc (and make more C lib conforming) and your inputs are timely and very welcome!

PR #639, opened 2020-05-23
[jvm] Put all @Override in place

andreoss:
Add `@Override` on overriden methods, remove `public` modifier in interfaces, also remove boxing and use diamond where applicable.

### DIFF
--- a/src/vm/jvm/runtime/org/raku/nqp/io/AsyncFileHandle.java
+++ b/src/vm/jvm/runtime/org/raku/nqp/io/AsyncFileHandle.java
@@ -53,6 +53,7 @@ public class AsyncFileHandle implements IIOClosable, IIOEncodable, IIOAsyncReada
         }
     }
 
+    @Override
     public void close(ThreadContext tc) {
         try {
             chan.close();
@@ -61,6 +62,7 @@ public class AsyncFileHandle implements IIOClosable, IIOEncodable, IIOAsyncReada
         }
     }
 
+    @Override
     public void setEncoding(ThreadContext tc, Charset cs) {
         enc = cs.newEncoder();
         dec = cs.newDecoder();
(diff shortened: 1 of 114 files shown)


usev6:
I've run a spectest with this changes and didn't see any fallout.

I'm not a Java developer, so I don't dare to do a final assessment here. That being said, adding those ```@Override``` and using the lambda syntax for Threads look good (and uncontroversial) to me. (I'd hit the merge button for those changes.)

What I'm not completely sure about are the removals of a couple of ```static/public/static public``` keywords and the removal of boxing (https://github.com/Raku/nqp/pull/639/commits/8dc6af179d3b468f49e60e4ece2e8af3c3ee24aa). I guess, those are fine too, but don't fully understand the effects of these changes and I don't know why we had the old version in the first place. Also I remember https://github.com/Raku/nqp/pull/507, that was submitted to clean up the Java code. It ended up being reverted and earned an unhappy comment from @jnthn about mainly making ```git blame``` less useful (which I can confirm ;)). I can't imagine that the current PR would get the same assessment, but still ...

I want to add that I'm very happy to see efforts to improve/cleanup/revive the JVM backend. Thanks a lot, @andreoss .

andreoss:
> What I'm not completely sure about are the removals of a couple of `static/public/static public` keywords and the removal of boxing ([8dc6af1](https://github.com/Raku/nqp/commit/8dc6af179d3b468f49e60e4ece2e8af3c3ee24aa)). 

In the cases I found the modifiers were redundant, such as `public` for an interface method, and `static` for an inner interface. 
And while auto-boxing has some drawbacks, I doubt the explicit boxing was indented here (or may be some very early rakudo/nqp versions had to run on prior-1.5 JVM).
Same goes with diamond operator and lambdas.

PR #643, opened 2020-06-16
[JVM] Make gather work with last/next/redo in loop

usev6:
This helps with a long standing bug on the JVM backend:
https://github.com/rakudo/rakudo/issues/3634

If last/next/redo were called within an extra block (marked as
'immediate' or 'immediate_static'), that call was translated to
a direct JVM level call. In this case the produced UnwindException
never hit it's handler.

I'm not sure this is a good way to fix the problem. But at least
it points at the relevant code.

Another caveat: If last/next/redo is used with a label, we still
get an uncaught UnwindException.

### DIFF
--- a/src/vm/jvm/QAST/Compiler.nqp
+++ b/src/vm/jvm/QAST/Compiler.nqp
@@ -4080,6 +4080,10 @@ class QAST::CompilerJAST {
                 'Void', $TYPE_TC, $TYPE_CR ));
             $*JMETH.append(JAST::Instruction.new( :op('astore'), 'cf' ));
 
+            # Add redo label. We might need to handle a control exception.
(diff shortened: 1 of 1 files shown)


niner (approved):

usev6:
@niner Thanks for looking.
I'm still not really sure this is a good way to fix the problem. I'll see if I'm able to find a way to refine the code so that it also helps with labeled last/next/redo.
Anyway, there is always the option to revert, so I'll merge the PR now.

niner:
That's what I thought. It doesn't solve everything and it doesn't look pretty, but it's a clear improvement. If in doubt, I'd always take that and add lots of commentary of the requirements and shortfalls of this code, so when in the future someone takes a step back to look at a more general solution, the details may be more clear than they were initially.

PR #647, opened 2020-07-25
Register a plain HLL::Compiler object under the 'default' name

patrickbkr:
This allows the code in NQP to use functionality in the HLL Compiler.
Before that wasn't possible as it wasn't known which name the subclassed
Compiler object would have.

Use that generic name to access it from NQP code. In particular in the
logic that writes the `--profile` output.

This fixes rakudo/rakudo#3824

This is has received close to no testing. (But a simple `raku --profile -e "say 'hi'"` did work.) Also I don't know if it's a good way to expose the HLL::Compiler object in this way.

### DIFF
--- a/src/HLL/Compiler.nqp
+++ b/src/HLL/Compiler.nqp
@@ -844,3 +844,4 @@ class HLL::Compiler does HLL::Backend::Default {
 
 my $compiler := HLL::Compiler.new();
 $compiler.language($compiler.backend.name);
+nqp::bindcomp('default', $compiler);
--- a/src/vm/moar/HLL/Backend.nqp
+++ b/src/vm/moar/HLL/Backend.nqp
@@ -699,7 +699,7 @@ class HLL::Backend::MoarVM {
         my str $template;
 
         if !$want_json && !$want_sql {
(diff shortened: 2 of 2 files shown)


vrurg:
I wonder why _HLL_, not _default_ or _current_? Just curious.

vrurg:
Ok, first saw the PR, then your comment in the issue.

Basically, the current situation doesn't leave us other options but to use `Compiler`. We can introduce another class for the purpose, but to get the object of the class we would need something like `nqp::getcomp` again.

Besides, the `Compiler` already deals with command line options. `nqp-home` is not much away from them.

A stylistic workaround to the situation would be to have an attribute `$!environment` or alike on the `Compiler` which would be responsible for the environment the compiler is being invoked in. The term nicely cover the command line options, which are currently defined on the `Compiler`, and `nqp-home`, and whatever other information we may need. But this approach is quite arguable too.

patrickbkr:
Then I'll go with the current approach.

There was no specific reason I chose 'HLL'. It was simply the first name that came to mind. Looking at your ideas I like 'default' better. I changed it accordingly.

PR #648, opened 2020-07-25
[JVM] Make osname available in nqp::backendconfig

usev6:
This key is available on the other backends, and there are a couple
of special cases in Rakudo's code base because the operating system
name had to be taken from nqp::jvmgetproperties(){'os.name'}.

### DIFF
--- a/src/HLL/Compiler.nqp
+++ b/src/HLL/Compiler.nqp
@@ -817,13 +817,12 @@ class HLL::Compiler does HLL::Backend::Default {
     method nqp-home() {
         if !$!nqp-home {
             # Determine NQP directory
+            my $config := nqp::backendconfig();
+            my $sep := $config<osname> eq 'MSWin32' ?? '\\' !! '/';
 #?if jvm
-            my $sep := nqp::atkey(nqp::jvmgetproperties,'os.name') eq 'MSWin32' ?? '\\' !! '/';
             my $execname := nqp::atkey(nqp::jvmgetproperties,'nqp.execname') // '';
 #?endif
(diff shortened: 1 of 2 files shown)


usev6:
Probably I should have mentioned this in the commit message: The result returned from nqp::backendconfig() might differ from the result that nqp::jvmgetproperties() gave. (Most notably the case differs.) But on the plus side, we get the same results for MoarVM and JVM now:
```
freebsd$ ./rakudo-j -e 'use nqp; say "Old output on FreeBSD on JVM: ", nqp::jvmgetproperties(){"os.name"}; say "New output on FreeBSD on JVM: ", nqp::backendconfig(){"osname"}'
Old output on FreeBSD on JVM: FreeBSD
New output on FreeBSD on JVM: freebsd
```

```
freebsd$ ./rakudo-m -e 'use nqp; say "Output on FreeBSD on MoarVM: ", nqp::backendconfig(){"osname"}'
Output on FreeBSD on MoarVM: freebsd
```

```
linux$ $ ./rakudo-j -e 'use nqp; say "Old output on Linux on JVM: ", nqp::jvmgetproperties(){"os.name"}; say "New output on Linux on JVM: ", nqp::backendconfig(){"osname"}'
Old output on Linux on JVM: Linux
New output on Linux on JVM: linux
```

```
linux$ raku -e 'use nqp; say "Output on Linux on MoarVM: ", nqp::backendconfig(){"osname"}'
Output on Linux on MoarVM: linux
```

I'm pretty sure this won't cause any trouble.

PR #649, opened 2020-07-25
[JVM] Implement op 'execname'

usev6:
This is a bit hacky, but it should enable us to use nqp::execname
in Rakudo's code base uniformly on all backends.

Currently we use nqp::jvmgetproperties(){nqp.execname} for NQP
and nqp::jvmgetproperties(){perl6.execname} for Rakudo.
Those properties are passed to the JVM via the runner scripts nqp-j
and rakudo-j, respectively.

I've also added support for raku.execname, so perl6.execname could
be easily removed in the future.

### DIFF
--- a/src/HLL/Compiler.nqp
+++ b/src/HLL/Compiler.nqp
@@ -820,6 +820,7 @@ class HLL::Compiler does HLL::Backend::Default {
             my $config := nqp::backendconfig();
             my $sep := $config<osname> eq 'MSWin32' ?? '\\' !! '/';
 #?if jvm
+            # TODO could be replaced by nqp::execname() after the next bootstrap for JVM
             my $execname := nqp::atkey(nqp::jvmgetproperties,'nqp.execname') // '';
 #?endif
 #?if !jvm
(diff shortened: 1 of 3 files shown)


patrickbkr:
Is there a reason we chose `raku.execname`, `nqp.execname` and so forth instead of just `execname`? On MoarVM the commandline parameter is called `--execname`.

Maybe we can rename the property on JVM to also simply be `execname`?

usev6:
(I force pushed to resolve a conflict in src/HLL/Compiler.nqp.)

@patrickbkr I wondered about that too, but I thought it would be strange/atypical to name a property just ```execname```. I searched a bit now and it seems like there is no clear rule or convention for naming properties. But using hierarchicle names and grouping them under a prefix (like ```myapp```) seems to be a common approach.

There are some system properties related to Java like ```java.home```, ```java.vendor```, ```java.version```, other system properties like ```os.name```, ```os.version``` related to the OS. So ```raku.execname``` and ```nqp.execname``` look more fitting to me -- even if it's things get a bit more ugly under the hood.

patrickbkr:
@usev6 LGTM

PR #650, opened 2020-07-30
Support for typechecking against Raku roles

vrurg:
If a `NQPClassHOW` class get a Raku role mixin then typechecking fails
for complex cases like parameterized roles. By incorporating
`role_typecheck_list` provided by role's HOW we fix the problem.

NQP, Rakudo, and roast tests are passing.


vrurg:
Related is rakudo/rakudo#3834

PR #653, opened 2020-08-03
Add a HLL::SysConfig class

patrickbkr:
This class acts as a central point where configuration information is stored. The class should only ever be accessed via `nqp::gethllsym('default', 'SysConfig')`. This then allows the class to be overridden by a subclass.

### DIFF
--- a/src/HLL/Compiler.nqp
+++ b/src/HLL/Compiler.nqp
@@ -8,13 +8,11 @@ class HLL::Compiler does HLL::Backend::Default {
     has @!cmdoptions;
     has $!compiler_progname;
     has $!language;
-    has %!config;
     has $!user_progname;
     has @!cli-arguments;
     has %!cli-options;
     has $!backend;
     has $!save_ctx;
-    has $!nqp-home;
 
     method BUILD() {
         # Backend is set to the default one, by default.
@@ -32,7 +30,6 @@ class HLL::Compiler does HLL::Backend::Default {
         ~ ' confprog=s'
 #?endif
         );
-        %!config     := nqp::hash();
     }
 
     method backend(*@value) {
@@ -55,7 +52,7 @@ class HLL::Compiler does HLL::Backend::Default {
         nqp::getcomp($name);
     }
 
-    method config() { %!config };
+    method sysconfig() { nqp::gethllsym('default', 'SysConfig').nqp-build-config }
 
     method autoprint($value) {
         self.interactive_result($value)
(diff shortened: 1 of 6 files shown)


src/HLL/Compiler.nqp
```diff
@@ -55,7 +52,7 @@ class HLL::Compiler does HLL::Backend::Default {
         nqp::getcomp($name);
     }
 
-    method config() { %!config };
+    method config() { nqp::gethllsym('default', 'SysConfig').nqp-build-config }
```
vrurg, line 55: A very minor note: the method name confused me into thinking that the config would be rebuilt upon every `.config()` call.

src/HLL/SysConfig.nqp
```diff
@@ -0,0 +1,34 @@
+class HLL::SysConfig {
+    my %build-config := nqp::hash();
+    my $nqp-home;
+
+    method nqp-build-config() { %build-config };
+
+    method nqp-home() {
+        if !$nqp-home {
+            # Determine NQP directory
+            my $sep := nqp::backendconfig<osname> eq 'MSWin32' ?? '\\' !! '/';
```
vrurg, line 10: I think Configure.pl must put the separator into the hll-config when we generate it. Too often a line like this one appears in many places across the code.

vrurg: Moreover, SysConfig can provide basic filespec routines for the rest of the code. We can think about it later.

src/HLL/Compiler.nqp
vrurg: I doubt if we actually need the `config()` method with `nqp::gethllsym('default', 'SysConfig')` available. It's current name suggest that it's compiler's configuration.

I would consider something like `sysconfig`, or `sys-config`, or `system-config`. Perhaps the first one is the best due to its consonance with with the hllsym and class name.

patrickbkr: Done.

src/HLL/SysConfig.nqp
patrickbkr: I didn't retrieve it from the config but calculate it in place. That makes the logic not depend on the configuration. That doesn't force us to initialize the configuration before retrieving the separator and also we don't have to worry about the configuration to either be the NQP or Rakudo one.

patrickbkr:
I'm going to merge this and the respective Rakudo PR now, so we'll have some days of testing left before the next release. If problems show up we'll have some time to react. If this is somehow fatally broken we can revert.

PR #657, opened 2020-08-17
More tests for NQP hash behaviour.

nwc10:
In particular, this adds tests for iterators, which previously were completely untested, including the ability to delete at the current iterator position, which was never directly tested before.

It also tests exceptions thrown on iterators. I was kind of assuming that the MoarVM behaviour was considered "correct" (or "canonical" here. In particular, the slightly "interesting" behaviour that you have to advance the freshly created iterator before you can use it.

Turns out that

- JS doesn't have any exceptions implemented yet
- JVM doesn't throw exceptions before the iterator is advanced to the first element
- JVM and MoarVM differ in the text of the exception thrown when the iterator is exhausted

1. Are NQP exceptions supposed to be part of the NQP language "spec"?
2. Even if "no", for this case would it be useful if the JVM backend did something more careful, as what it throws currently is `java.util.NoSuchElementException` which I assume can also be thrown for (some types of) regular hash **lookup**.

Everything passes, thanks to the magic of `TODO`.

### DIFF
--- a/t/nqp/108-vmhash.t
+++ b/t/nqp/108-vmhash.t
@@ -73,10 +73,11 @@ try {
 todo('Exceptions for iterators NYI on js', 2)
     if $backend eq 'js';
 ok($msg ne "", 'iterator throws after end');
-# Is it "useful" that these two are different?
-is($msg, $backend eq 'jvm' ?? 'java.util.NoSuchElementException'
-       !! 'Iteration past end of iterator',
-       'iterator throws correct exception after end');
+# As per comments on https://github.com/Raku/nqp/pull/657
+# the JVM backend should be changed to throw the same error text
+todo('Error message not yet consistent on the JVM', 1)
+    if $backend eq 'jvm';
+is($msg, 'Iteration past end of iterator', 'iterator throws correct exception after end');
 
 $hash<a> := 1;
 is(nqp::elems($hash), 2, 'hash has 2 elements once more');
--- a/t/nqp/108-vmhash.t
+++ b/t/nqp/108-vmhash.t
@@ -1,4 +1,7 @@
-plan(10);
+plan(28);
+
+my $backend := nqp::getcomp('nqp').backend.name;
+
 class NQPHashClass is repr('VMHash') {
 }
 
@@ -14,7 +17,7 @@ my $clone := nqp::clone($hash);
 $clone<b> := 3;
 is($hash<b>, 2, "changing clone doesn't modify orignal");
 is($clone<a>, 1, 'clone has value taken from orginal');
-is($clone<b>, 3, "clone has a value of it's own");
+is($clone<b>, 3, 'clone has a value of its own');
 
 is(nqp::elems($hash), 2, 'nqp::elems on hash');
 is(0+$hash, 2, 'numifing a hash');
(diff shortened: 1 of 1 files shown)


jnthn:
I'd say iteration past the end wants a consistent exception (with a consistent text), so that part seems good to me. Can argue whether implementations should be required to throw in the case that you didn't yet move to the next element; probably it's not particularly costly to require that of them.

nwc10:
Both CI tests that failed were here:

```
2020-08-17T14:04:09.1416478Z # Failed test 'Both precompilation runs resulted in the same checksum'
2020-08-17T14:04:09.1417704Z # at t/02-rakudo/reproducible-builds.t line 32
2020-08-17T14:04:09.1427419Z # expected: '9ED11DBB9F84CEF7637471093529B0805E60C139'
2020-08-17T14:04:09.1428349Z #      got: '75F28F356D96FEDE665B6A4E275529C43035F19F'
2020-08-17T14:04:09.1494163Z Unexpected named argument 'bin' passed
2020-08-17T14:04:09.1494854Z   in block  at t/02-rakudo/reproducible-builds.t line 37
2020-08-17T14:04:09.1495646Z   in block <unit> at t/02-rakudo/reproducible-builds.t line 32
2020-08-17T14:04:09.1496004Z 
2020-08-17T14:04:09.1559388Z t/02-rakudo/reproducible-builds.t ............................... 
2020-08-17T14:04:09.1560254Z not ok 1 - Both precompilation runs resulted in the same checksum
2020-08-17T14:04:09.1560733Z Dubious, test returned 1 (wstat 256, 0x100)
2020-08-17T14:04:09.1561102Z Failed 1/1 subtests 
```

I don't *think* that that has anything do with this change. Does that test just flap sometimes?

nwc10:
> I'd say iteration past the end wants a consistent exception (with a consistent text), so that part seems good to me. Can argue whether implementations should be required to throw in the case that you didn't yet move to the next element; probably it's not particularly costly to require that of them.

I could sort of look at it the other way - given that there (have to) exist iterators that are "before the first element", and these therefore cannot point to any element in the hash, and that it is possible to dereference them, what other behaviour should they have on a dereference? Return `undef` for both key and value? If the underlying VM provides a `first`/`next`/`next`/... API for hashes, then that is going to have to be slightly special cased to map it to how NQP wants it, so it seems as easy to make that code throw as to make it return something else.

(Implying that the patch as written is correct to declare that the missing exception is a `todo`, not a `skip`)

(And isn't yet correct, in that it accepts what the JVM currently offers, `java.util.NoSuchElementException`, instead of also treating that as a `todo`)

jnthn:
> I don't think that that has anything do with this change. Does that test just flap sometimes?

Didn't notice it doing so before. Did a re-run, we'll find out.

> I could sort of look at it the other way - given that there (have to) exist iterators that are "before the first element", and these therefore cannot point to any element in the hash, and that it is possible to dereference them, what other behaviour should they have on a dereference? Return undef for both key and value? If the underlying VM provides a first/next/next/... API for hashes, then that is going to have to be slightly special cased to map it to how NQP wants it, so it seems as easy to make that code throw as to make it return something else.
> 
> (Implying that the patch as written is correct to declare that the missing exception is a todo, not a skip)

OK, agree. :-)

> (And isn't yet correct, in that it accepts what the JVM currently offers, java.util.NoSuchElementException, instead of also treating that as a todo)

Yes; if you're interested to make it that way please do.

jnthn (approved):

PR #666, opened 2020-09-14
Move build logic to separate method so it can be called in subclasses

patrickbkr:
This is a support PR for [rakudo/rakudo#3913](https://github.com/rakudo/rakudo/pull/3913).
Do not merge. The PR might possibly not be needed in the end.


patrickbkr:
The PR turned out to actually be needed.

PR #669, opened 2020-09-25
Move all tests for streaming decoder to t/nqp

usev6:
This proposal is related to https://github.com/Raku/nqp/issues/668: I plan to add more tests to the new test file t/nqp/116-streaming-decoder.t, and it feels clearer to have the tests in one common test file.

I might have missed a reason for having backend-specific test files (for the decoder tests). If so, it would be nice if someone could point that out to me.


lizmat (approved):

jnthn:
> I might have missed a reason for having backend-specific test files (for the decoder tests). If so, it would be nice if someone could point that out to me.

Only because the API was initially only implemented in full on MoarVM. If it's being ported more thoroughly, then moving the test file makes sense (this just saved todo'ing stuff all over in the past).

jnthn (approved):

PR #670, opened 2020-09-25
[JVM] Use singleton object for VMnull

usev6:
When I introduced the VMNull for the JVM backend (https://github.com/Raku/nqp/pull/605), @pmurias++ suggested to use a null singleton. I'm not sure my change matches what he aimed at, though.


niner (approved):
I'm pretty sure this is what pmurias had in mind as it reflects what MoarVM is doing.

jnthn (approved):
Mildly different factoring from how I think it is in MoarVM, but it makes sense to me.

PR #673, opened 2020-09-26
Please do not merge: Add more tests for streaming decoder and fixes for JVM

usev6:
... a lot of them skipped for the JVM backend.

Adding these tests might introduce some redundancy (both, within
the new tests and in combination with the existing tests). But
each block exercises a specific functionality and shows existing
problems on the JVM backend (not covered by tests before). So some
redundancy should be ok.


usev6:
This is related to https://github.com/Raku/nqp/issues/668

usev6:
Interestingly the last commit makes the test for https://github.com/Raku/old-issue-tracker/issues/6278 (in S32-io/io-cathandle.t) fail on the JVM backend. That test is failing for MoarVM, too. IMHO it's ok to accept this regression on the JVM backend for now.

usev6:
I'm closing this one in favour of https://github.com/Raku/nqp/pull/674.

I only understood in retrospect what I got wrong on my first try. So, sorry for the noise.

PR #674, opened 2020-09-28
Improve streaming decoder on the JVM backend

usev6:
Fixes https://github.com/Raku/nqp/issues/668

The added tests might introduce some redundancy (both, within
the new tests and in combination with the existing tests). But
each block exercises a specific functionality and shows existing
problems on the JVM backend (not covered by tests before).


usev6:
If no-one objects, I'll go ahead and merge this PR tomorrow.

lizmat (approved):

PR #681, opened 2020-11-02
Change NQP to Not Quite Paradise

lizmat:
Inspired by https://www.reddit.com/r/rakulang/comments/jmvnf2/renaming_nqp/

### DIFF
--- a/README.pod
+++ b/README.pod
@@ -1,11 +1,8 @@
 =pod
 
-=head1 NQP - Not Quite Perl
+=head1 NQP - Not Quite Paradise
 
-NQP is Copyright (C) 2009-2020 by The Perl Foundation.  See F<LICENSE>
-for licensing details.
-
-This is "Not Quite Perl" -- a lightweight Raku-like environment for
+This is "Not Quite Paradise" -- a lightweight Raku-like environment for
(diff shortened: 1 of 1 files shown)


ugexe:
I never understood the desire to rename NQP... "Not Quite Perl" is as accurate a description now as it was 10 years ago.

raiph:
For @ugexe, @vrurg, @coke, and anyone else who agrees or disagrees with @ugexe's earlier comment:

> "Not Quite Perl" is as accurate a description now as it was 10 years ago.

I agree.

As you'll see if you read my comment, I did not suggest in my reddit post that NQP stops being Not Quite Perl.

lizmat:
@raiph having "NQP" being a Junction, is, in my mind almost just as bad as having "Raku" being an alias for Perl 6.  But the feelings are clear.  And it's not a big beef of mine, I also just wanted to throw it out there.  And am throwing it back in again now.

xfix:
Not Quite Perl is a fine name in my opinion, I don't think Not Quite Perl should be renamed (other than maybe NQP, as in, removing acronym entirely). Just because the language was renamed from Perl 6 to Raku doesn't mean NQP has to change.

PR #683, opened 2020-11-21
[JVM] Align behaviour for NQP hash to MoarVM

usev6:
The JVM backend now throws for certain operations on hashes in
the same way as MoarVM:

* Iterating past the end of the iterator dies with "Iteration past
  end of iterator" instead of throwing a NoSuchElementException.

* Before starting to iterate and after reaching the end of the
  iterator, trying to access key or value gives a sane error message.


usev6:
BTW, https://github.com/Raku/nqp/pull/657 has more context with regard to the thrown exception.

niner (approved):

jnthn (approved):

PR #684, opened 2020-11-28
Make mixins concurrency safe

niner:
Multiple threads creating mixins on the same type of object could step on each
other's toes as the mixin cache was not protected against concurrent access.
Instead of adding locks, we now use the VM's parameterization mechanism, which
is already concurrency safe and used for mixins in Rakudo.

Fixes "Iteration past end of iterator" when running:
raku -e '(^10000).race.map({EVAL "q[];\n"}).sink'

### DIFF
--- a/src/how/NQPClassHOW.nqp
+++ b/src/how/NQPClassHOW.nqp
@@ -2,6 +2,24 @@
 # that Raku needs, but it's sufficient for NQP. Supports methods, attributes,
 # role composition, inheritance (single and multiple) and introspection.
 
+knowhow NQPMixinCacheHOW {
+    method new(:$name = '<anon>') {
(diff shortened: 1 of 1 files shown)


src/how/NQPClassHOW.nqp
```diff
@@ -2,6 +2,26 @@
 # that Raku needs, but it's sufficient for NQP. Supports methods, attributes,
 # role composition, inheritance (single and multiple) and introspection.
 
+knowhow NQPMixinCacheHOW {
+    method new(:$name = '<anon>') {
+        my $obj := nqp::create(self);
+        #$obj.BUILD(:name($name));
```
MasterDuke17, line 8: This line isn’t needed?

PR #688, opened 2020-12-12
floating point atime should be compared with integer atime, not mtime.

nwc10:
Similarly ctime with ctime, not mtime.

This appears to be a copy-paste thinko that dates back to when the test was
originally added in 2016 by commit 37862cd52d63eb61:
    Add subsecond file time ops stat_time and lstat_time


lizmat (approved):

ilmari:
Yeah, that was clearly a copy-pasto on my part, sorrly.

jnthn (approved):

PR #692, opened 2021-01-03
Minor README corrections

JJ:
Here and there.


niner (changes requested):
I think the addition is misleading. Rakudo can be installed into a different --prefix than MoarVM. You just need to give the appropriate --with-moar option to Configure.pl.

coke:
```
unless you use the `--with-moar`.
```

Should probably read

```
unless you use the option `--with-moar`.
```

PR #694, opened 2021-01-23
Stub some Archetypes methods expected by Rakudo

Kaiepi:
Exceptions can come about from these being missing at times. No need for attributes; we don't have these luxuries here.

Edit: this is premature. Need a minute to test more.

### DIFF
--- a/src/how/Archetypes.nqp
+++ b/src/how/Archetypes.nqp
@@ -21,7 +21,7 @@ my knowhow Archetypes {
     # Can this be composed (either with flattening composition, or used
     # as a mixin)?
     has $!composable;
-    
+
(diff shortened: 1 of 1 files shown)


niner:
Odd. How can adding a couple of methods lead to NULL pointer exceptions on JVM?

vrurg:
@niner It looks like an earlier commit caused the breakage, not this PR.

Just for linking purpose: this should fix rakudo/rakudo#4162

niner:
@vrurg thanks for clarification. I merged it now. Came across the same issue this weekend. E.g. this prevented one from having a signature like `sub foo(QAST::Block $block)`

PR #701, opened 2021-03-04
Give NQPParametricRoleHOW a role_typecheck_list method

Kaiepi:
This allows Raku parametric roles to do nqp ones (e.g. weird HOWs).

### DIFF
--- a/src/how/NQPParametricRoleHOW.nqp
+++ b/src/how/NQPParametricRoleHOW.nqp
@@ -19,6 +19,7 @@ knowhow NQPParametricRoleHOW {
 
     # Other roles that this one does.
     has @!roles;
+    has @!role_typecheck_list;
 
     # Have we been composed?
     has $!composed;
@@ -105,6 +106,14 @@ knowhow NQPParametricRoleHOW {
 
     # Compose the role. Beyond this point, no changes are allowed.
     method compose($obj) {
+        @!role_typecheck_list := my @rtl;
+        for @!roles {
+            @rtl.push: $_;
+            for $_.HOW.role_typecheck_list($_) {
+                @rtl.push: $_;
+            }
+        }
+
         $!composed := 1;
         nqp::settypecache($obj, [$obj.WHAT]);
         nqp::setmethcache($obj, {});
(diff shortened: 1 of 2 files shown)


src/how/NQPParametricRoleHOW.nqp
```diff
@@ -105,6 +106,14 @@ knowhow NQPParametricRoleHOW {
 
     # Compose the role. Beyond this point, no changes are allowed.
     method compose($obj) {
+        @!role_typecheck_list := my @rtl;
```
MasterDuke17, line 109: Why are you using this `my @rtl` instead of just pushing directly to `@!role_typecheck_list`?

Kaiepi: At this point, `@!role_typecheck_list` hasn't been bound yet. Should probably be bound sooner to avoid `role_typecheck_list` returning a `VMNull` before composition.

MasterDuke17: Ah, then what about `@!role_typecheck_list := nqp::list();` up in `BUILD()`?

Kaiepi: That works.

vrurg:
Does it unlock `class Foo does QAST::Children { }` achievement? I'd like to better understand what problem do we solve here.

Kaiepi:
In my `Data::Record` module, record types are managed by `MetamodelX::RecordHOW` and `MetamodelX::RecordTemplateHOW` metaclasses. The former has a list of fields associated with its instances. For list and tuple types, that's fine, but for map types, that needs to be coerced to a `Map` at one point or another. By making them parametric metaroles instead, I could ensure fields are typed as the types they're associated with like, but since `MetamodelX::RecordTemplateHOW` does `Metamodel::Naming` and Rakudo expects `role_typecheck_list` to exist on roles done by roles, that throws.

What's the problem with `class Foo does QAST::Children { }`? The type composes and I can create instances of it and set their children OK.

vrurg (approved):

Kaiepi:
Can this be merged?

vrurg:
> What's the problem with `class Foo does QAST::Children { }`?

It's hard for me to recollect what I was experimenting with back then, but most likely I was trying not a class but `role Foo does QAST::Children`. 

Anyway, thanks for reminding. I think it's time to merge it.

PR #702, opened 2021-03-04
Use faster `while nqp::elems(@)` vs `while @`

MasterDuke17:
These lists probably aren't very long, but might as well use the more
efficient form.

### DIFF
--- a/src/HLL/Actions.nqp
+++ b/src/HLL/Actions.nqp
@@ -200,7 +200,7 @@ class HLL::Actions {
         }
         if $lastlit gt '' { @parts.push(QAST::SVal.new( :value($lastlit) )); }
         my $ast := @parts ?? @parts.shift !! QAST::SVal.new( :value('') );
-        while @parts {
+        while nqp::elems(@parts) {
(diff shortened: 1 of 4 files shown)


lizmat:
OOC, what is the difference in codegen between the two?  Maybe the codegen should be changed?

MasterDuke17:
> OOC, what is the difference in codegen between the two?

It's just 
```
    │     - QAST::Op(while)  @a { say(nqp::shift(@a)) }
    │       - QAST::Var(local __lowered_lex_1) 
    │       - QAST::Stmts  { say(nqp::shift(@a)) }
```
vs
```
    │     - QAST::Op(while)  nqp::elems(@a) { say(nqp::shift(@a)) }
    │       - QAST::Op(elems)  nqp::elems(@a)
    │         - QAST::Var(local __lowered_lex_1) 
    │       - QAST::Stmts  { say(nqp::shift(@a)) }
```

> Maybe the codegen should be changed?

Heh, yeah, just had that same thought over lunch. I'll give it a look.

MasterDuke17:
Will close if https://github.com/Raku/nqp/pull/703 is merged.

PR #703, opened 2021-03-04
Optimize truth checks of arrays in whiles and ifs

MasterDuke17:
E.g., `while @a { ... }` and `if @a { ... }` are now converted into
`while nqp::elems(@a) { ... }` and `if nqp::elems(@a) { ... }`.

This micro benchmark:
```raku
my @a;
my int $i := 0;
while $i++ < 10_000_000 {
	nqp::push(@a, $i);
}
my $b;
my num $s := nqp::time_n;
$b := nqp::shift(@a) while @a;
say(nqp::sub_n(nqp::time_n, $s));
say($b)
```
used to take 0.08s, now it takes 0.04s.

### DIFF
--- a/src/NQP/Optimizer.nqp
+++ b/src/NQP/Optimizer.nqp
@@ -246,6 +246,23 @@ class NQP::Optimizer {
                 $op[1].blocktype($orig);
             }
         }
+        elsif $opname eq 'while' {
+            my $conditions := $op[0];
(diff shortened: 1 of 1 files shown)


lizmat:
While I like the result, I can't help but wonder if this should live in the optimizer.  @jnthn ?

MasterDuke17:
> While I like the result, I can't help but wonder if this should live in the optimizer.

Where else do you think it should go? I'm not sure we have enough information/context in Grammar or Actions.

niner (approved):

PR #704, opened 2021-03-18
Eliminate the ops sec_n, asec_n and sech_h, which are no longer used.

nwc10:
These hyperbolic functions are not provided directly by libc, so have to be
implemented in terms of functions which are. There's no mathematical
difference in expressing the same identities in the Rakudo setting (it's
going to dispatched as the same IEEE operations) but it's better to do it
once there instead of once per backend.

Rakudo stopped using these ops in the 2021.02 release; now now remove them
from NQP and MoarVM.

The JS implementation of sech this commit removes is curious - it's
different from the JVM implementation, MoarVM's C implementation, all the
other JS implementations I can find online, and the "textbook" definition:

    sech 𝑥 = 1 / cosh 𝑥

The relevant commits don't give any clues.

However, I *believe* it to be mathematically equivalent - after a lot of
head scratching and dead ends, the identities that one needs to re-arrange
from it to the canonical definition above are that cosh 2𝑥 = cosh²𝑥 + sinh²𝑥
and cosh²𝑥 = 1 + sinh²𝑥. Hence *why* - it makes two cosh calls whereas the
"obvious" formula uses only one, *and* it has to special case ±∞. I *think*
that it might produce "better" results (ie - not just 0.0) for inputs whose
output /could/ be correctly expressed as an IEEE subnormal. But that's the
only reason I can see for doing this; I can't find it described anywhere,
and it seems a lot of work for a corner case of a corner case, which isn't
even going to be mathematically rigorous. And the fact that it needs to
special case Inf to avoid evaluating 0.0/0.0 and hence returning NaN makes
me wonder if it (still) produces similar results for some very large inputs.

Really, if you care about extreme results, you should be calculating in
symbolic algebra anyway...

Of course, if I'm wrong on this, then this approach ought to be used in the
Rakudo setting (along with the explanation of why it's objectively better.)

### DIFF
--- a/docs/ops.markdown
+++ b/docs/ops.markdown
@@ -489,14 +489,11 @@ The opcodes are grouped into the following categories:
 ## [Trigonometric](#trig)
 
 [acos](#acos) |
-[asec](#asec) |
 [asin](#asin) |
 [atan](#atan) |
 [atan2](#atan2) |
 [cos](#cos) |
 [cosh](#cosh) |
-[sec](#sec) |
-[sech](#sech) |
 [sin](#sin) |
 [sinh](#sinh) |
 [tan](#tan) |
@@ -3276,11 +3273,6 @@ an integral number of seconds, `_n` returns a fractional amount.
 
 Arccosine.
 
-## asec
-* `asec_n(num $n --> num)`
-
-Arcsecant.
-
 ## asin
 * `asin_n(num $n --> num)`
 
@@ -3305,16 +3297,6 @@ Cosine.
 * `cosh_n(num $n --> num))`
 
 Hyperbolic cosine.
-
-## sec
-* `sec_n(num $n --> num))`
-
-Secant.
-
-## sech
-* `sech_n(num $n --> num))`
-
-Hyperbolic secant.
 
 ## sin
 * `sin_n(num $n --> num))`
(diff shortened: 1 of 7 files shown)


MasterDuke17:
Would it make sense to also re-bootstrap?

nwc10:
That's a good question. We're not using these ops in any of the bootstrap bytecode so we don't **need** to. (Strictly, because of this I don't think that we need to even if we re-used the op IDs for different new OPs). Re-bootstrapping feels like churn, so I'd be tempted to say "no", because there's no real need to currently. I'd prefer to wait until there's actual need to re-bootstrap. But I might have missed something in my reasoning.

We don't have a policy for this written down, do we?

MasterDuke17:
I don't think there's a written policy. But I agree that it doesn't seem necessary for this change (and they're ops that someone is unlikely to suddenly try and use in the build), so can avoid bloating the repo with more binaries. Also, I may have a go at changing `nqp::time_(i|n)` re https://github.com/rakudo/rakudo/issues/3620#issuecomment-613561305 in the nearish future and that would require a rebootstrap.

MasterDuke17 (approved):

jnthn (approved):

PR #707, opened 2021-03-30
Remove Travis badge

patrickbkr:
Travis has been disabled since the beginning of January already.

Someone should also have a look at the GitHub configuration. I suspect, the hook for Travis is still active.

Ping @Altai-man. Do you have access to the GitHub permissions and could have a look?


JJ:
It's not going to be activated if there's no .travis.yml present.

Altai-man:
Is the second one the one I should delete?
2021-03-30-200454_2560x1440_scrot

patrickbkr:
@Altai-man I suspect so. Even though I have to admit, that I have never configured Travis, so can't tell for sure. But I'd have no idea what else that hook would be used for.

PR #708, opened 2021-04-01
[JVM] Add op 'nativecallinvoke'

usev6:
For now, it's identical to 'nativecall'. The op was introduced for
MoarVM with https://github.com/Raku/nqp/commit/28e0650f1c and it is
used in Rakudo's lib/NativeCall.rakumod.


niner (approved):
Ah, yes, this makes sense! On MoarVM nativecallinvoke doesn't decont. Only the nativecall desugar does. The generated body of native functions already deconts the arguments where that's necessary, so we don't have to do it at the lower levels. But deconting again also doesn't hurt and is reasonably fast. So we'll even still get a speedup despite it.

PR #710, opened 2021-04-05
[JVM] Make nqp::unbox_i on NativeCall object work

usev6:
It is supposed to report whether the library was loaded or not. This
is a port of https://github.com/MoarVM/MoarVM/commit/1b0950e3e9 and
is needed since https://github.com/rakudo/rakudo/commit/9ef7a7e9dd.

nine++ gave the hint for the concrete implementation on the JVM.

Please note, that the usage of nqp::unbox_i($!call) in
lib/NativeCall.rakudmod don't work yet. They result in a
NullPointerException -- but that's probably a different bug. At least
the following code works now:

  use nqp;
  use NativeCall;
  my class native_callsite is repr('NativeCall') { };
  my $bar = native_callsite.new;
  say nqp::unbox_i($bar);  ## 0

### DIFF
--- a/src/vm/jvm/runtime/org/raku/nqp/runtime/Ops.java
+++ b/src/vm/jvm/runtime/org/raku/nqp/runtime/Ops.java
@@ -2633,7 +2633,7 @@ public final class Ops {
                     retval = attr_st.REPR.allocate(tc, attr_st);
                     for (Field field : retval.getClass().getDeclaredFields()) {
                         try {
-                            if (field.getType().isAssignableFrom(tc.native_j.getClass())) {
+                            if (tc.native_j == null || field.getType().isAssignableFrom(tc.native_j.getClass())) {
                                 field.set(retval, tc.native_j);
                                 break;
                             }
--- a/src/vm/jvm/runtime/org/raku/nqp/runtime/Ops.java
+++ b/src/vm/jvm/runtime/org/raku/nqp/runtime/Ops.java
@@ -2554,7 +2554,7 @@ public final class Ops {
             return obj.get_attribute_boxed(tc, decont(ch, tc), name, STable.NO_HINT);
         }
         catch (P6OpaqueBaseInstance.BadReferenceRuntimeException badRef) {
-            SixModelObject retval = null;
+            SixModelObject retval = createNull(tc);
             obj.get_attribute_native(tc, decont(ch, tc), name, STable.NO_HINT);
             if (tc.native_type == ThreadContext.NATIVE_INT) {
                 retval = box_i((long)tc.native_i, tc.curFrame.codeRef.staticInfo.compUnit.hllConfig.intBoxType, tc);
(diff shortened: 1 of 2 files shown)


usev6:
This as far as I got for now. With some bandaid on Rakudo's side this results in
```
t/04-nativecall/00-misc.t ................... No subtests run
t/04-nativecall/01-argless.t ................ Failed 13/14 subtests
t/04-nativecall/02-simple-args.t ............ ok
t/04-nativecall/03-simple-returns.t ......... ok
t/04-nativecall/04-pointers.t ............... Failed 11/22 subtests
t/04-nativecall/05-arrays.t ................. Failed 46/46 subtests
t/04-nativecall/06-struct.t ................. Failed 18/34 subtests
t/04-nativecall/07-writebarrier.t ........... Failed 7/7 subtests
t/04-nativecall/08-callbacks.t .............. ok
t/04-nativecall/09-nativecast.t ............. ok
t/04-nativecall/10-cglobals.t ............... ok
t/04-nativecall/11-cpp.t .................... Failed 21/22 subtests
t/04-nativecall/12-sizeof.t ................. ok
t/04-nativecall/13-cpp-mangling.t ........... ok
t/04-nativecall/13-union.t .................. ok
t/04-nativecall/14-rw-attrs.t ............... ok
t/04-nativecall/15-rw-args.t ................ ok
t/04-nativecall/16-rt125408.t ............... ok
t/04-nativecall/16-rt125729.t ............... ok
t/04-nativecall/17-libnames.t ............... ok
t/04-nativecall/18-routine-sig-sanity.t ..... ok
t/04-nativecall/19-function-pointers.t ...... ok
t/04-nativecall/20-concurrent.t ............. ok
t/04-nativecall/21-callback-other-thread.t .. Failed 8/9 subtests
t/04-nativecall/22-method.t ................. Failed 2/2 subtests
t/04-nativecall/23-incomplete-types.t ....... Failed 6/6 subtests
```

As far as I can see most of the remaining problems are related to the same problem: The block in https://github.com/rakudo/rakudo/blob/995d39c03b/lib/NativeCall.rakumod#L584 gets called with the wrong arguments (or from the wrong place). This needs further investigation. My guess is that this is not a problem in NQP, but in Rakudo.

One thing to note: I've found a way to make t/04-nativecall/04-pointers.t pass, too. It fails with ```java.lang.UnsupportedOperationException: This pointer is opaque: const@0x82e200b78``` when trying to dereference a pointer to an array in https://github.com/rakudo/rakudo/blob/995d39c03b/t/04-nativecall/04-pointers.t#L34
With the following change the test passes:
```
--- a/src/vm/jvm/runtime/org/raku/nqp/sixmodel/reprs/CPointerInstance.java
+++ b/src/vm/jvm/runtime/org/raku/nqp/sixmodel/reprs/CPointerInstance.java
@@ -11,7 +11,7 @@ public class CPointerInstance extends SixModelObject {
     @Override
     public void set_int(ThreadContext tc, long value) {
         this.pointer = (value > 0)
-            ? Pointer.createConstant(value)
+            ? new Pointer(value)
             : null;
     }
 
```

But the [JNA documentation](https://github.com/java-native-access/jna/blob/a0641dd928/src/com/sun/jna/Pointer.java#L75) says ```Don't use this unless you know what you're doing.```. I don't really understand the consequences, so I'm not trying to change this right now. Maybe I'll create a dedicated issue for this problem.

usev6:
I somehow forgot about this PR. I still think it improves the situation wrt NativeCall. Does anyone want to look?

Otherwise I'll probably merge this myself (after some days). (There's always the option to revert.)

jonathanstowe (approved):

PR #712, opened 2021-04-08
Remove no longer needed $!slang override in Braid

MasterDuke17:
It's been long enough that modules should have been fixed by now.

Not entirely sure how to test this, but NQP builds ok and passes `make m-test`, Rakudo builds ok and passes `make m-test m-spectest`, and both Slang::SQL and Slang::Tuxic install ok and pass their (albeit quite limited) tests.


lizmat:
in the raku-community-modules there are some other `Slang` modules if I recall correctly.  Not sure if they've bitrotted too far or not, but maybe something to check?

MasterDuke17:
I did a search for "slang" on modules.raku.org and all the ones found (Slang::Kazu, Slang::Subscripts, Slang::AltTernary, Slang::Piersing, Slang::Mosdef, Slang::Nogil, Slang::Roman) passed their tests.

PR #715, opened 2021-04-13
No need to stringify the float twice in the `sprintf` implementation.

nwc10:
This is a small oversight introduced in Han 2019 by commit afeef2b971c28ef4:
    First stab at making num use explicit


nwc10:
Pants. A glaring typo. Now fixed.

MasterDuke17 (approved):

lizmat (approved):

jnthn (approved):

PR #716, opened 2021-04-20
Accurately calculate pow_n(10.0, $exp) for floating point sprintf formatting

nwc10:
The problem we have is that for floating point formatting, sprintf needs accurate powers of 10, and nqp::num_n(1e1, $exp) is a call to the libc pow() function, and **that** may not always be accurate, with the result that we generate incorrect formatting. pow() is accurate (for all positive powers) on the little endian Linux platforms I tested but on ppc64 and sparc64, `nqp::say(nqp::pow_n(10, 210))` returns `1.0000000000000001e+210` (which I think is an error of 1 bit in the last place of the mantissa) However, MinGW seems to be much worse, with errors of 4 or 5 bits.

The obvious approach would be to implement the algorithm from MoarVM's `pow_i`, but in NQP in nums:

    while (exp) {
        if (exp & 1)
            result *= base;
        exp >>= 1;
        base *= base;
    }

however, **this** doesn't work because floating point rounding errors from **some** multiplications happen to propagate, causing higher powers to be inaccurate. (and less accurate than the pow() function). Unlike pow, IEEE basic arithmetic is defined to be bit-perfect correct, but that definition includes rounding, and that can't be avoided.

After some head scratching, it turns out that for all powers of 10, there exist **some** smaller powers of 10 that can be multiplied together to calculate them, without rounding errors, but not in any pattern that can be expressed algorithmicly. However, we can generate a lookup table that stores which smaller powers to multiply, and if we ensure that one is always a "small" power which we can construct accurately as integers, then we can iterate our way to any power of 10.

Hopefully this approach resolves the outstanding test failures that block moarvm/moarvm#1385

### DIFF
--- a/tools/generate-sprintf-pow10-table.raku
+++ b/tools/generate-sprintf-pow10-table.raku
@@ -1,0 +1,164 @@
+use v6;
+
+# The problem we have is that for floating point formatting, sprintf needs
+# accurate powers of 10, and nqp::num_n(1e1, $exp) is a call to the libc pow()
+# function, and *that* may not always be accurate, with the result that we
+# generate incorrect formatting. pow() is accurate (for all positive powers)
+# on the little endian Linux platforms I tested but on ppc64 and sparc64,
+# nqp::say(nqp::pow_n(10, 210)) returns 1.0000000000000001e+210
+# (which I think is an error of 1 bit in the last place of the mantissa)
+# However, MinGW seems to be much worse, with errors of 4 or 5 bits.
+#
+# The obvious approach would be to implement the algorithm from MoarVM's pow_i,
+# but in NQP in nums:
+#     while (exp) {
+#         if (exp & 1)
+#             result *= base;
+#         exp >>= 1;
+#         base *= base;
+#     }
+#
+# however, *this* doesn't work because floating point rounding errors from
+# *some* multiplications happen to propagate, causing higher powers to be
+# inaccurate. (and less accurate than the pow() function). Unlike pow, IEEE
+# basic arithmetic is defined to be bit-perfect correct, but that definition
+# includes rounding, and that can't be avoided.
+#
+# After some head scratching, it turns out that for all powers of 10, there
+# exist *some* smaller powers of 10 that can be multiplied together to calculate
+# them, without rounding errors, but not in any pattern that can be expressed
+# algorithmicly. However, we can generate a lookup table that stores which
+# smaller powers to multiply, and if we ensure that one is always a "small"
+# power which we can construct accurately as integers, then we can iterate our
+# way to any power of 10.
+#
+# This code generates and verifies that table. The table is pasted into
+# sprintf.nqp, where there is an NQP implementation of pow10()
+#
+# See https://github.com/MoarVM/MoarVM/pull/1385
+
+use nqp;
+
+# MoarVM decimal parsing is accurate. The platform's pow function might not be:
+my @truth = [1e0, 1e1, 1e2, 1e3, 1e4, 1e5, 1e6, 1e7, 1e8, 1e9, 1e10, 1e11, 1e12,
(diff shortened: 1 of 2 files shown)


MasterDuke17:
Looks like there is an implementation (https://github.com/Raku/nqp/blob/master/src/vm/jvm/runtime/org/raku/nqp/runtime/Ops.java#L6714-L6716), but it's not mapped to `coerce_in`. Probably just need to add `QAST::OperationsJAST.map_classlib_core_op('coerce_in', $TYPE_OPS, 'coerce_i2n', [$RT_NUM], $RT_INT);` around https://github.com/Raku/nqp/blob/master/src/vm/jvm/QAST/Compiler.nqp#L2964

nwc10:
Thanks. "Will fix" and "this is why we test"

And the nutty bit is that I'd spotted that functions with **those** names existed in the JVM, before I found the proper MoarVM OP names, and that the JVM functions existed but were unused, but the implication of this didn't click.

MasterDuke17:
Although, I don't believe you need the explicit `coerce_in`, `my int $pow := 3; my num $val := nqp::pow_i(10, $pow); say($val)` works fine on the jvm backend.

nwc10:
> Although, I don't believe you need the explicit `coerce_in`, `my int $pow := 3; my num $val := nqp::pow_i(10, $pow); say($val)` works fine on the jvm backend.

I think I know what happened. That array population loop had been implemented with `nqp::pow_n`, which meant that I needed a call to `coerce_in` for the second argument. Then I thought "this is daft - the whole point of the rest of this code is because we don't trust the C librarly `pow` - so why am I using it here (when I think that it should be OK, but I should probably add a comment)". Then I realised that it's all value that `nqq::pow_i` can calculate, so I re-ordered it to call that, and moved the `coerce_in` outwards.

I didn't think to try just deleting it.

Thanks for spotting this and figuring out the best solution.

Fixed branch pushed.

nwc10:
@niner - is the initialisation code in the `class` block inside the `module` block early enough to count as `BEGIN` time? Or do I need an explicit `BEGIN` block? I don't know how to figure out whether the arrays as serialised, or the initialisation code gets run each time.

And oh my, that syntax is seductively useful. "Hmm, I think I need an array of native `num`s"...

        class Nums is repr('VMArray') is array_type(num) {
        }

Done!

nwc10:
@patrickbkr I think that this branch (built against MoarVM master - it doesn't need Ryū) will fix the failures mentioned on MoarVM/MoarVM#1385 on MinGW

patrickbkr:
@nwc10 I confirm this fixes the later two failures in 06-sprintf.t (but not the first one, which can be fixed by adapting the test itself). Can this be merged?

lizmat:
To be merged after the 2021.4 release, unless @Altai-man thinks it is ok to merge before.

patrickbkr:
After the release. :tada: Can we merge?

tools/generate-sprintf-pow10-table.raku
```diff
@@ -0,0 +1,164 @@
...
+# way to any power of 10.
+#
+# This code generates and verifies that table. The table is pasted into
+# sprintf.nqp, where there is an NQP implementation of pow10()
+#
+# See https://github.com/MoarVM/MoarVM/pull/1385
+
+use nqp;
+
+# MoarVM decimal parsing is accurate. The platform's pow function might not be:
```
niner, line 42: Silly question: if MoarVM's decimal parsing is accurate and can generate the true values we're looking for, why not just hard code this table in the actual sprintf code?

nwc10: It works out at 2.4K of storage, most of which isn't needed by most programs, whereas I figured that about 440 bytes plus some other lookup code was about 25% of the size. To me, that felt a better trade off.

PR #717, opened 2021-04-25
Revert "Lazy allocation of HashMap for  VMHash storage on JVM."

usev6:
This reverts commit f7c396af6b3bf6f3260c4d8c7507284a86ca53dc.

It turned out that we had at least one place where we didn't check if
the storage of a VMHash was still VMHashInstance.EMPTY_MAP:

  https://github.com/Raku/nqp/blob/f1469fad3d/src/vm/jvm/runtime/org/raku/nqp/sixmodel/reprs/VMHash.java#L53

Since VMHashInstance.EMPTY_MAP was shared between different instances
of VMHash, putting key/value pairs into storage affected all VMHashes
that didn't get their dedicated storage, yet.

If I'm not mistaken this was the underlying reason for

  my constant $foo = nqp::hash();

not working correctly in the setting on the JVM backend. It also
explains strange problems with %DEPRECATIONS suddenly (after calling
nqp::deserialize) containing key/value pairs from other objects -- as
decribed in https://github.com/rakudo/rakudo/issues/3158.

Now for the actual fix: An alternative would have been to add a check
for VMHashInstance.EMPTY_MAP in deserialize_finish (see link above).
But since storage is a public field of each VMHashInstance, IMHO this
would be error prone, since each user would have to know about
VMHashInstance.EMPTY_MAP and check for it. So I didn't went that road.

Conflicts solved in:
    src/vm/jvm/runtime/org/raku/nqp/sixmodel/reprs/VMHashInstance.java

### DIFF
--- a/src/vm/jvm/runtime/org/raku/nqp/sixmodel/reprs/VMHash.java
+++ b/src/vm/jvm/runtime/org/raku/nqp/sixmodel/reprs/VMHash.java
@@ -25,7 +25,7 @@ public class VMHash extends REPR {
     public SixModelObject allocate(ThreadContext tc, STable st) {
         VMHashInstance obj = new VMHashInstance();
         obj.st = st;
-        obj.storage = VMHashInstance.EMPTY_MAP;
+        obj.storage = new HashMap<String, SixModelObject>();
         return obj;
     }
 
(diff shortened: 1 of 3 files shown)


niner (approved):

niner:
Wow, nice find!

PR #720, opened 2021-05-08
[JVM] Deserialize native arrays as such

usev6:
There is a long standing issue with Rakudo on the JVM backend:
If one declares a native array (like 'my int @arr;') this comes
out as a non-native array after precompilation.
(Compare https://github.com/rakudo/rakudo/issues/1666.)

I think I figured out what happened: Both, native and non-native
arrays have the REPR 'VMArray'. They differ in native arrays having
a non-null REPRData in their STable. When an array is serialized, a
string representing the REPR is put into the string heap. During the
first pass of deserialization, this string is used to create a stub
object (of the given REPR). At this moment the REPRData is not yet
known -- so a plain (non-native) VMArray is created. Only during the
second pass of deserialization, the REPRData is added to the STable
(in deserialize_repr_data). But that's too late to change the object.

My patch tweaks the string that is put into the string heap, so that
the correct stub object can be created. If I'm not mistaken, this fake
REPR name isn't used in another place and it is corrected during the
first pass of deserialization. This doesn't feel clean, but it seems
to work.

### DIFF
--- a/src/vm/jvm/runtime/org/raku/nqp/sixmodel/reprs/VMArray.java
+++ b/src/vm/jvm/runtime/org/raku/nqp/sixmodel/reprs/VMArray.java
@@ -88,7 +88,43 @@ public class VMArray extends REPR {
     public SixModelObject deserialize_stub(ThreadContext tc, STable st) {
         SixModelObject obj;
         if (st.REPRData == null) {
-            obj = new VMArrayInstance();
+            // Either a real VMArray or REPRData not yet known.
+            switch (st.REPR.name) {
+            case "VMArray":
+                obj = new VMArrayInstance();
+                break;
+            case "VMArray_i8":
+                obj = new VMArrayInstance_i8();
+                break;
+            case "VMArray_u8":
+                obj = new VMArrayInstance_u8();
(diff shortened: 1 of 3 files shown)


niner (approved):
It's a hack alright. But as hacks go a rather benign one.

src/vm/jvm/runtime/org/raku/nqp/sixmodel/reprs/VMArray.java
```diff
@@ -88,7 +88,43 @@ public void compose(ThreadContext tc, STable st, SixModelObject repr_info) {
...
+                obj = new VMArrayInstance();
+                break;
+            case "VMArray_i8":
+                obj = new VMArrayInstance_i8();
+                break;
+            case "VMArray_u8":
+                obj = new VMArrayInstance_u8();
+                break;
+            case "VMArray_i16":
+                obj = new VMArrayInstance_i32();
```
usev6, line 103: I only now noticed, that this should be ```VMArrayInstance_i16()``` (same copy-pasto for the unsinged variant). I'll force push a correction.

PR #721, opened 2021-05-08
[JVM] Keep different REPR.name for native arrays

usev6:
It turned out that f3dc977906 doesn't really fix the usage of native
arrays in the core setting. For some reason during the compilation of
blib/CORE.d.setting.jar native array are deserialized twice.

The first time worked as expected, but since we did reset REPR.name
to VMArray, the second time went wrong just as it used to faile before
f3dc977906.

Since REPR.name is only used for serialization/deserialization and for
error reporting in two base classes, it seems to be acceptable to just
keep the suffixes (VMArray_i, VMArray_u32, etc.).


usev6:
I'm not happy with this PR anymore. This will give wrong results with ```nqp::reprname```. I'll come up with a different PR.

PR #722, opened 2021-05-14
[JVM] Keep correct REPR.name for native arrays

usev6:
It turned out that f3dc977 doesn't really fix the usage of native
arrays in the core setting. For some reason during the compilation of
blib/CORE.d.setting.jar native array are deserialized twice.

The first time worked as expected, but since we did reset REPR.name
to VMArray, the second time went wrong just as it used to fail before
f3dc977.

Since REPR.name is also used for nqp::reprname, it seems cleaner to
leave it untouched and add a different field (vmarray_name) to REPR
instead.

### DIFF
--- a/src/vm/jvm/runtime/org/raku/nqp/sixmodel/REPR.java
+++ b/src/vm/jvm/runtime/org/raku/nqp/sixmodel/REPR.java
@@ -22,6 +22,12 @@ public abstract class REPR {
     public String name;
 
     /**
+     * Name that is used to differentiate between native and non-native
+     * VMArrays during deserialization. Not used for other representations.
+     */
+    public String vmarray_name;
+
+    /**
      * Creates a new type object of this representation, and associates it
      * with the given HOW.
      */
--- a/src/vm/jvm/runtime/org/raku/nqp/sixmodel/REPR.java
+++ b/src/vm/jvm/runtime/org/raku/nqp/sixmodel/REPR.java
@@ -22,10 +22,12 @@ public abstract class REPR {
     public String name;
 
     /**
-     * Name that is used to differentiate between native and non-native
-     * VMArrays during deserialization. Not used for other representations.
+     * A name that is used to differentiate between different (sub)types that
+     * have the same representation otherwise.
+     * Can be left unset for most representations, but is used to detect
+     * native VMArrays during deserialization.
      */
-    public String vmarray_name;
+    public String subtype_name;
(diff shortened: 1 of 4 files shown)


usev6:
Spectest is running.

niner:
> Since REPR.name is also used for nqp::reprname, it seems cleaner to
> leave it untouched and add a different field (vmarray_name) to REPR
> instead.

Does it actually have to be a name? The required information could be 
communicated in a single byte. After all it's just a number of bits and a flag 
for whether it's signed or not. But then, that's probably the C programmer in 
me talking, always trying to be faster and more memory efficient. Don't even 
know if it would matter.

But at least let me bikeshed a little: since this additional field may be 
useful for other REPRs as well, a more generic name should be more 
appropriate. Maybe "subtype"?

usev6:
... and looked good (no additional failures).

usev6:
>> Since REPR.name is also used for nqp::reprname, it seems cleaner to leave it untouched and add a different field (vmarray_name) to REPR instead.

> Does it actually have to be a name? The required information could be communicated in a single byte. After all it's just a number of bits and a flag for whether it's signed or not. But then, that's probably the C programmer in me talking, always trying to be faster and more memory efficient. Don't even know if it would matter.

I spent one or two thoughts on that and had the feeling that using a string wouldn't have a noticeable impact on performance, but would make the code a tiny bit clearer (esp. since we already work with strings in REPRRegistry). But using an int (or even a byte) would probably do as well. Maybe just something like the following (I'll have to look where to define that):
```
int VMARRAY = 0;
int VMARRAY_I = 1;
int VMARRAY_N = 2;
int VMARRAY_S = 3;
int VMARRAY_I8 = 4;
int VMARRAY_U8 = 5;
...
```

>But at least let me bikeshed a little: since this additional field may be useful for other REPRs as well, a more generic name should be more appropriate. Maybe "subtype"?

Yes, allowing to use the additional fields for other things sounds good. I'm fine with "subtype". But if we stay with a string, maybe "subtype_name" would be acceptable, too?

niner:
> But if we stay with a string, maybe "subtype_name"
> would be acceptable, too?

Sure!

usev6:
I've changed the code to use "subtype_name" (but it's still a string). I still have a slight preference for using a string, but I'll happily change to a byte if you insist ;)

niner (approved):

PR #723, opened 2021-05-18
Use Win32 native API to get file changetime on JVM

MasterDuke17:
Windows doesn't support "unix:ctime" when using Files.getAttribute, so
instead try calling the Win32 native API. This required upgrading JNA
from 4.0.0 to 4.5.0 and adding the JNA platform jar.

Gets us one test closer to passing `make j-test` on Windows.


usev6 (approved):
I don't know if I'm really qualified to approve this, but the changes look sane to me. Since the new jna libraries seem to work fine (CI jobs passes) and it helps on Windows, I'm all for it :)

PR #724, opened 2021-05-22
Partially implement translate_newlines config in JVM backend's DecoderInstance

MasterDuke17:
Add a boolean member for it (defaults to false) and set it based on the
config hash if given.

Currently only takeAllChars uses it, because that's all that was needed
to get both t/nqp/019-file-opt.t and t/nqp/116-streaming-decoder.t to
pass on Windows with the JVM backend.

This will allow us to turn back on the JVM-on-Windows NQP test job in CI, which I accidentally did in #719, so after a rebase that PR should all pass.

### DIFF
--- a/src/vm/jvm/runtime/org/raku/nqp/sixmodel/reprs/DecoderInstance.java
+++ b/src/vm/jvm/runtime/org/raku/nqp/sixmodel/reprs/DecoderInstance.java
@@ -69,6 +69,10 @@ public class DecoderInstance extends SixModelObject {
         }
     }
 
+    public synchronized String maybe_translate_newlines(ThreadContext tc, String str) {
+        return translate_newlines ? str.replace("\r\n", "\n") : str;
+    }
+
     public synchronized String takeChars(ThreadContext tc, long chars, boolean eof) {
         ensureConfigured(tc);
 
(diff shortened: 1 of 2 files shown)


src/vm/jvm/runtime/org/raku/nqp/sixmodel/reprs/DecoderInstance.java
```diff
@@ -156,7 +159,8 @@ public synchronized String takeAllChars(ThreadContext tc) {
             decoder.flush(target);
             decoder.reset();
         }
-        return Normalizer.normalize(decodedBuffer(target), Normalizer.Form.NFC);
+        String normalized = Normalizer.normalize(decodedBuffer(target), Normalizer.Form.NFC);
+        return translate_newlines ? normalized.replace("\r\n", "\n") : normalized;
```
jnthn, line 163: I think this should happen in `takeChars` and `takeAvailableChars` too, probably by factoring out this line into a method. Otherwise it seems to be doing the right thing.

MasterDuke17: Cool, I'll give that a shot and see what happens.

MasterDuke17: Done. No change in tests.

jnthn (approved):

PR #725, opened 2021-06-05
Simplify loop in optimizer's incorporate_inner

MasterDuke17:
I don't know why this splice was originally commented out and not used,
but it works.


lizmat:
git blame:

```
f7a23cadd3 (Stefan Seifert       2019-04-27 16:49:48 +0200  72)                         for %to_add{$name} { nqp::push(@existing, $_) }
d062aa3ca9 (Jonathan Worthington 2014-04-03 23:59:21 +0200  73)                         #nqp::splice(@existing, $_.value, 0, 0);
```

PR #726, opened 2021-06-05
Micro-optimize the sift_down sub in sorted_keys

MasterDuke17:
Exit early if there are 0 or 1 keys to sort, also pull some variable
declarations out of a loop to reduce the number of BOOTInts allocated.

### DIFF
--- a/src/core/Hash.nqp
+++ b/src/core/Hash.nqp
@@ -15,10 +15,12 @@ sub sorted_keys($hash) {
 
     sub sift_down(@a, int $start, int $end) {
         my int $root := $start;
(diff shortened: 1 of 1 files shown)


lizmat:
In the `sort` routine in Rakudo, I've built in special casing for 2-element lists.  Would that make sense here as well?  It shouldn't have to make any difference for the 3+ elements case, if you change the `< 2` condition to `< 3` and check for `== 2` after that.

MasterDuke17:
Right, I thought about that and then forgot to give it a try. I’ll see if it makes a difference.

PR #730, opened 2021-06-20
Implement support for moar::hllincludes config variable

vrurg:
List of include directories is now provided by moar backend and respected by NQP build.

### DIFF
--- a/tools/templates/Makefile-backend-common.in
+++ b/tools/templates/Makefile-backend-common.in
@@ -188,4 +188,4 @@
 @backend_prefix@-qregex-test-loud: @bpm(BUILD_RUNNER)@
 	$(PROVE) -r -v --exec @nfpq(@bpm(BUILD_RUNNER)@)@ @nfp(t/qregex)@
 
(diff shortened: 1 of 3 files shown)


vrurg:
@MasterDuke17 do you have an idea why Windows build fails? Looks like it's something to do with libuv compatibility.

vrurg:
BTW, it is reproducible on my local Win7 VM with Visual Studio. Another VM successfully builds with GNU.

MasterDuke17:
Hm, not 100% sure why I think your PR was from a 3 month old NQP. But regardless, if you rebase up to HEAD of master, it should pull in a MoarVM bump that will fix this.

vrurg:
@MasterDuke17 It shouldn't be. I think the diff of PR would be much larger than, wouldn't it?

But I have rebased to pull in your latest bump. Let's see if it helps.

MasterDuke17:
Yeah, this was in my kludged-together Windows VM dev environment, so who knows what actually happened. Maybe I'd pulled commits, but not tags, so git describe was wrong? Anyway, looks good now, thanks.

PR #731, opened 2021-07-06
Add a test ensuring a Hash iterator misuse is not exploding

Altai-man:
Closes https://github.com/rakudo/rakudo/issues/1340 (aims to)

I am really not sure where the nqp integration tests should belong, but this appears like the most fitting place to me, please check those with wisdom.


t/nqp/108-vmhash.t
```diff
@@ -234,3 +234,16 @@ todo('Exceptions for iterators NYI on js', 2)
...
+    my $h := nqp::hash("a", 42);
+    my $i := nqp::iterator($h);
+    nqp::deletekey($h, "a");
+    nqp::shift($i);
+    CATCH {
+        nqp::getmessage($_);
+    }
+}
+
+ok($msg eq "", 'No SEGV on a misused hash iterator');
```
lizmat, line 249: I'm not seeing `$msg` being changed?  Is that just to create a valid `ok` test?  If that was the intent, you can also say:

`pass 'No SEGV on a misused hash iterator'`

Altai-man: Oh, my bad as I did not add `$msg = ...` bit into CATCH, huh.
Also, given it is not directly an exception, but a SEGV is maybe a better option, let's see...

Altai-man: There is apparently no `pass` in nqp, so just corrected the bit.

Altai-man:
The test failures seem to be flappers. :(

t/nqp/108-vmhash.t
japhb: Altai-man: It looks like the test actually checks for *both* no SEGV (by happening at all) and no exception (by leaving `$msg` unchanged).  Might be worth changing the test description to match that.

PR #732, opened 2021-07-17
Wrap QAST::Regex from qbuildsub in QAST::Stmts...

MasterDuke17:
and add a `:node` parameter to `qbuildsub()` and pass in `$/` when it's
called, this is then set as the `:node` argument to the new QAST::Stmts
call. The combination of these changes means that we now get a filename
set for more regex parts of a backtrace. E.g., `raku --ll-exception -e
'grammar G { token TOP { $<foo>=(<xxx>) } }; G.parse("x")'` now reports
'-e' instead of '\<unknown\>' for the backtrace line before the 'TOP'.

This results in one passing TODO in t/spec/S32-exceptions/misc2.t and
one new fail in t/spec/S02-lexical-conventions/comments.t (gets an
`X::Syntax::Comment::Embedded` instead of an `X::Syntax::Confused`), but
I think the new exception is a better one and roast should be changed.

### DIFF
--- a/src/QRegex/P6Regex/Actions.nqp
+++ b/src/QRegex/P6Regex/Actions.nqp
@@ -7,7 +7,7 @@ class QRegex::P6Regex::Actions is HLL::Actions {
             :compilation_mode(0),
             :pre_deserialize($*W.load_dependency_tasks()),
             :post_deserialize($*W.fixup_tasks()),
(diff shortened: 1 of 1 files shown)


MasterDuke17:
Closes https://github.com/Raku/old-issue-tracker/issues/3960

jnthn (approved):

PR #740, opened 2021-10-04
Make JVM tests pass after new-disp merge

MasterDuke17:
The JVM doesn't implement nqp::dispatch, so move any tests with that to
the MoarVM tests directory.

The JVM also hasn't implemented hllize via dispatchers, so just skip
that test on the JVM.

### DIFF
--- a/t/moar/54-dispatch-qast.t
+++ b/t/moar/54-dispatch-qast.t
@@ -1,0 +1,47 @@
+use QAST;
+
+plan(1);
+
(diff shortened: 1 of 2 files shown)


usev6:
Looks good to me.

jnthn:
Oops, sorry; I did test the JVM build during `new-disp` at certain times (and tried to conditionalize everything so as to keep it working), but forgot to make another check of it in the excitement to merge.

PR #746, opened 2021-10-31
[JVM] Deal with invalid input in decodertakeline

usev6:
As long as we have lines of correctly encoded input, decodertakeline
is supposed to return a valid line -- even if there is invalid data
(for the given encoding) later in the input. It should die only if
it reaches the invalid input on the line it wants to return.

A concrete example of this behaviour is exercised in the method
!read-dependencies in Rakudo's CompUnit::PrecompilationUnit::File:
The precomp files we handle there have some leading lines of text,
followed by an empty line and then there is binary data.
In the method we call IO::Handle.get repeatedly -- which uses
nqp::decodertakeline under the hood. The old code needed a dirty
workaround to not explode at this point. Now the situation is
handled more correctly.


usev6:
As an early warning: I plan to merge this before the next release ;)

lizmat:
Better earlier than later!

PR #747, opened 2021-11-11
Eliminate use of nqp::attrinited

jnthn:
This is a follow-up to https://github.com/rakudo/rakudo/pull/4633 that also removes the use of `nqp::attrinited` in NQP; it cannot be merged until the Rakudo PR has been merged. This does not do a rebootstrap, which would be the final step needed before MoarVM can drop support, as further changes to object representation configuration are also planned.


jnthn:
Also, the tests ain't gonna pass before the Rakudo one is merged either.

PR #750, opened 2021-11-30
- Allows the proper use of --no-optimize in --moar-option

Xliff:
Previously, even if --moar-option="--no-optimize" was used, the "--optimize" switch was still added to MoarVM's invocation of ./Configure.pl.

This patch addresses that issue.

### DIFF
--- a/tools/lib/NQP/Config/NQP.pm
+++ b/tools/lib/NQP/Config/NQP.pm
@@ -331,7 +331,7 @@ sub gen_moar {
     my $force_rebuild = $options->{'force-rebuild'};
     my $has_gen_moar  = defined $gen_moar;
     my @opts          = @{ $options->{'moar-option'} || [] };
-    push @opts, "--optimize";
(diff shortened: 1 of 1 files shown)


tools/lib/NQP/Config/NQP.pm
```diff
@@ -331,7 +331,7 @@ sub gen_moar {
     my $force_rebuild = $options->{'force-rebuild'};
     my $has_gen_moar  = defined $gen_moar;
     my @opts          = @{ $options->{'moar-option'} || [] };
-    push @opts, "--optimize";
+    push @opts, "--optimize" unless grep('--no-optimize', @{ $options->{'moar-option'} });
```
niner, line 334: I don't think, this works as intended. It will still always set --optimize:
```
> perl -E 'say grep("d", qw(b a c));'
bac
> perl -E 'say grep({$_ eq "d"} qw(b a c));'

> perl -E 'say grep({$_ eq "a"} qw(b a c));'
a
```

niner (changes requested):

PR #754, opened 2022-01-01
Replace flatnamed arg flag by flat | named

niner:
No need to have a separate flag when the combination of the existing flags
means the same thing. Will free up the flat for other use in MoarVM.


niner:
Has been merged manually a week ago

PR #756, opened 2022-01-09
More preparations for fixing unsigned handling

niner:
This PR includes the final round of fixes for unsigned handling that can be split off from the main branch. These commits are independent of changes in the other layers, i.e. they are still compatible with the master branches of the other repositories.

The fixes are mostly about using unsigned data types when handling unsigned data in MAST code, using unsigned ops for handling unsigned arrays and preparation for giving unsigneds their own objprimspec (10).


usev6:
As this breaks the build on the JVM backend: Do you think there is an alternative to porting your changes to the other backends? Since I didn't really follow your work on fixing/improving the handling of unsigned ints I don't have an idea how much of a change that would require.

niner:
> As this breaks the build on the JVM backend: Do you think there is an
> alternative to porting your changes to the other backends?

No, I don't see another way.

> Since I didn't
> really follow your work on fixing/improving the handling of unsigned ints I
> don't have an idea how much of a change that would require.

Well, the changes are not actually that difficult. The hard part was really 
finding out the actual scope of the problem, i.e. how much changes we actually 
need to arrive at a new stable state. And bootstrapping issues of course.

Most of the new code is really just copy & paste with s/_i/_u/ or something 
equivalent. It's really mostly passing down information about whether to use 
the get_int or get_uint REPR function in the end.

PR #757, opened 2022-01-16
Fix handling of unsigned native integers

niner:
These are the final fixes which accompany the changes to how unsigned integers are handled in MoarVM.


niner:
Merged it manually

PR #764, opened 2022-02-08
Remove some not-needed `try`s, or convert to `nqp::can` + the method call

MasterDuke17:
NQP builds ok and passes `make m-test j-test` and Rakudo builds ok and passes `make m-test m-spectest`.

The one commit I'm not 100% sure about is e1d60fb534b047fdd807fc1601a3090720a2eb72. I've never seen that error happen while taking profiles, so I'm gambling that it's safe to remove the `try`.


MasterDuke17:
Too annoyed by the profiler commits, just removing them entirely.

PR #767, opened 2022-03-04
Fix all returned native integers getting treated as signed

niner:
Use the new ops for dispatching with unsigned result and returning native
unsinged integers on MoarVM


lizmat (approved):

lizmat:
is this ready for merge?

PR #768, opened 2022-03-22
Some potential MAST bytecode writing optimizations

MasterDuke17:
Reduce the max memory used when writing MAST bytecode by pre-sizing the buffer to the estimated final size when writing a frame. Additionally, then each frame is written to the file individually instead of all of them being concatenated in memory and written at once. Given the prevalence and performance of modern SSDs, lots of small writes isn't as slow as it used to be with spinning disks. Before, building Rakudo would case the creation of multiple large VMArrays (up to >9m elements) that would grow by only very small amounts very many times. Now only a couple are created and they grow by a few large requests. Below are some statistics gathered with /usr/bin/time for building the CORE.c and CORE.e settings.

```
          rss        time
 
 master: 
   e      212,942    3.43
   c      1,767,033  58.05
   
 patch:
   e      212,577    3.54
   c      1,758,348  57.94
```


MasterDuke17:
Requires a small patch to MoarVM to add the `write_uint16_at` method to `MAST::Bytecode`.

lizmat:
For my understanding: this is just for memory usage, not for wallclock?  As it looks like it is slower than master?

niner (commented):
The changes look ok (except for the excessive line length). However I can't help but dislike them. The proposed code looks quite fiddly. There is redundancy between the write_* method and the indx increment. Easy to get those out of sync. I wonder if the MAST::Bytecode class could keep track of the current index for us? Maybe with an additional set of methods like add_uint32 or the like?

MasterDuke17:
Yeah, if the idea in general is approved I can give it a cleanup pass. I'll give that a shot and then update this PR.

niner (approved):

PR #772, opened 2022-05-17
Give nqp named anon methods

Kaiepi:
Allows for the following pattern:

```raku
nqp::getstaticcode(
    anon method introspective() {
        nqp::die('nice callstack!')
    });
```

Which would be *very* useful in Rakudo's bootstrapping, for instance. Many of Rakudo's core methods lack names on their code objects, which tends to be rather confusing for newcomers introspecting methods.

Unsure of how to add a test for this.


Kaiepi:
This should demonstrate that this behaves:

```raku
class WithNamedAnon {
    method named_anon() {
        nqp::getstaticcode(anon method named_anon() { })
    }
}

my &named_anon := WithNamedAnon.named_anon();
say(nqp::getcodename(&named_anon) eq 'named_anon');
say(!(named_anon(WithNamedAnon) =:= &named_anon));
```

PR #776, opened 2022-07-24
Make some enhancements to atomic integers

Kaiepi:
- Give nqp native refs to allow for usage of atomic ops in nqp.
- Introduce a semaphore-like `Counter` that performs comparably.
- I'm getting tired of writing every atomic op twice in Rakudo. Maybe it's time to give the JVM and JS backend atomicops?
- Let Rakudo building at all be the test of native refs, since they will be required at that point in time soon enough.
- `Counter` needs sanity tests in its own right. Swapping out `Semaphore` and its `acquire/release` for a `Counter` and its `want_pred`/`drop_succ` passes OK.
- Takes the place of https://github.com/Raku/nqp/pull/774.

### DIFF
--- a/src/core/Counter.nqp
+++ b/src/core/Counter.nqp
@@ -1,0 +1,87 @@
+# This adds bounds to a native atomic uint, giving a few means of handling when
+# those are reached via succ/pred operations:
+#
+# - drop_pred and drop_succ perform said operation only if valid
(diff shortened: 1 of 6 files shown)


lizmat:
It would be really nice if someone could continue the work of Kaiepi

@patrickbkr @ab5tract @timo @niner

PR #779, opened 2022-09-19
[JVM] Improve handling of unicode numbers

usev6:


usev6:
I didn't see any additional spectest breakage from this changes.

usev6:
I'm tempted to merge this even without getting another pair of eyes.

Sure, the addition to ```unipropcode``` and ```getuniprop_str``` look quite ad hoc and are nothing in comparison to what MoarVM does there. (Even though I picked the same numbers as MoarVM for ```UNIPROP_NUMERIC_VALUE_NUMERATOR``` and ```UNIPROP_NUMERIC_VALUE_DENOMINATOR```.)

But on the other hand the current situation (described in https://github.com/rakudo/rakudo/issues/4760) is clearly LTA. The usage of ```nqp::getuniprop_str``` in Rakudo's ```method numish``` (in https://github.com/rakudo/rakudo/blob/30a9d3a40a/src/Perl6/Actions.nqp#L8526-L8527) is quite central -- and the JVM backend can do better there. I'd hope the small patch wouldn't hinder any future improvements in that area.

lizmat (approved):

usev6:
@lizmat thanks for looking!

PR #780, opened 2022-10-07
Add unsigned comparison ops

MasterDuke17:


lizmat (approved):
Looks good to me, glad to see JVM also covered!

MasterDuke17:
Oops, looks like as of Java 8 there’s a Long.compareUnsigned(). I’ll switch to using that later today.

lizmat:
Taking that as my cue to do the Rakudo part, right?

PR #787, opened 2022-12-09
Better backtrace

vrurg:
The rather anonymous backtraces of NQP code were always pissing me off! Bugs reported as they are originated at the beginning of a block; no support for `#line` in concatenated files which may consist of tens of thousands of lines with no way to point out into the actual original file where the problematic piece of code came from...

Not anymore! ;)


MasterDuke17:
You might need to re-bootstrap since you’re changing the compiler. Does it work locally without?

vrurg:
Locally I have the same problem. I considered re-bootstrapping too and tried it today morning. No success.

Some mental modeling makes me think that there is something wrong about the filename and that checksum is calculated over different paths. Plan to check this out later today.

PR #790, opened 2023-01-03
bump jna lib for jvm

hughsimpson:
jna introduced M1 support with v5.7.0 (cf https://github.com/java-native-access/jna/blob/master/CHANGES.md#release-570). Bumping to latest release version for various subsequent fixes and because new shiny is nice


hughsimpson:
Looks like there are some failures here so probably not ready for a review yet. May be the jline dependency...

vrurg:
Failures are expected for JVMs because the backend still lacks support for RakuAST and fails on related tests. The only suspicious one is win_mvm_relocatable, but it is barely related and I think it was simply a fluctuation. Let's see what re-run would end up with.

hughsimpson:
@jdv any chance of a look at this one?

hughsimpson:
Not much point in leaving this open, very outdated anyway

PR #794, opened 2023-01-21
[JVM] Add some constants DISP_*

usev6:
At least DISP_ONLYSTAR is referenced by Rakudo, so it makes sense to define them similarly to src/vm/moar/QAST/QASTOperationsMAST.nqp.


usev6:
Since adding the constants should be rather harmless and since this is one of the problems that break the JVM build for Rakudo, I plan to hit the merge button in a couple of days -- unless someone objects.

niner (approved):

PR #797, opened 2023-02-05
[JVM] Add op `jvmgetunicodeversion`

usev6:


lizmat:
I'm actually going to create a separate PR taking Ralph Mellor's idea of a Unicode class.  Having this op on the JVM backend will be helpful for that.

PR #809, opened 2023-09-10
[JVM] Work around error for back-references

usev6:
... with aliases. For some reasons calling &has_aliased_names fails with an external error: https://github.com/Raku/nqp/issues/808.

The special-cased code for the JVM backend mimics the checks that are otherwise done in a single conditional (which includes a call to has_aliased_names).


usev6:
I'm inclined to merge this, since this helps with a problem on the JVM backend and shouldn't affect the other backends.

I'll wait for some more days, though.

PR #811, opened 2023-10-17
[JVM] Look at type check mode of checked value

usev6:
The old code could fail with values that had a different type check mode from the type that they are checked against. A simple example is the code from https://github.com/rakudo/rakudo/issues/2714:

  class C { };
  role R is C { };
  say R ~~ C;

That one still said False on the JVM backend, but works with this patch.

AFAICS this happened, because we kindly set the type check mode to 1 when setting up the role:
https://github.com/rakudo/rakudo/blob/55952f679a/src/Perl6/Metamodel/ParametricRoleGroupHOW.nqp#L269
But then we ignored that value, because we only looked at the type check mode of the type to check against (C).


usev6:
Since this only affects the JVM backend, I'll merge it now.

PR #815, opened 2024-03-18
Don't (re)compile P5QREGEX every time test are run

MasterDuke17:


jdv:
Note this was reverted because it broke the 2024.03 release.  With this change "make -j $num" fails in nqp which happens to be what the release process uses:  https://github.com/rakudo/rakudo/blob/main/tools/releasable/Akefile#L264.

PR #821, opened 2024-04-29
Draft: Add nqp::getnextcuid

ab5tract:
This is a mechanism for HLLs to rationally supply CUIDs at runtime without fear of collision.

I'm presenting it here for discussion as to its utility before I bother with JVM support and a better commit message.

### DIFF
--- a/docs/ops.markdown
+++ b/docs/ops.markdown
@@ -579,7 +579,9 @@ The opcodes are grouped into the following categories:
 [locallifetime](#locallifetime) |
 [setcodename](#setcodename) |
 [setdebugtypename](#setdebugtypename) |
-[takeclosure](#takeclosure)
+[takeclosure](#takeclosure) |
+[getnextcuid](#getnextcuid)
+
 
 # <a id="arithmetic"></a> Arithmetic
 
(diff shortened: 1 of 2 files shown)


niner:
What's the advantage over `use QAST; QAST::Block.next-uid`? And what's the use-case?

ab5tract:
AFAIK we don't expose `QAST` to userland like we do with `nqp`, but I could be wrong here.

```
./rakudo-m -MQAST -e1
===SORRY!=== Error while compiling -e
Could not find QAST in:
    ...
at -e:1
```

I don't have any particular use cases in mind (anymore). I opened the PR because the work is already done and I figured that it would be at least worth discussing.

Maybe this will enable something that myself and others haven't thought of yet?

lizmat:
Wouldn't a use case be `Inline::Perl5` an API to create CUIDs for its own usage, that are guaranteed not to interfere with Rakudo itself?

niner:
QAST is an NQP module, so the correction version is:
```
> rakudo -e 'use QAST:from<NQP>; say QAST::Block.next-cuid'
7
```

Inline::Perl5 doesn't need cuids. It needs to be able to export INIT phasers to its user. Cuids are just an implementation detail necessary for using the $*W.add_phaser implementation detail. It will need neither once we have a real solution in place.

PR #826, opened 2024-09-20
Sort files by name before hashing into source-digest

timo:
This makes the nqp build more reproducible (depending on the filesystem, and order of actions on folders, the ordering of files can change from time to time; reprotest from reproducible-builds can actually turn a file order jumbler on for different build attempts)

There is still one source of changes under reprotest which is the absolute path of the source files ends up in the debug symbols literally.


niner (approved):

timo:
it turns out using BASE_DIR for the templates in the makefile where i put the NQP_LIB dir instead was wrong, the things it compiles there are only for the time before installing, and what gets installed has different sources.

lizmat:
@timo so this PR needs more work, or is it mergeable ?

timo:
mergeable after a quick re-review

PR #827, opened 2024-11-11
Add is-interactive to HLL::Compiler

ab5tract:
This is a first-pass at allowing for refusing to start the REPL when there is no TTY available.

However, my preferred approach would be to nuke this concept of interactive vs non-interactive and replace with the following options for 'repl-mode':

    - tty
    - process
    - disabled

`disabled` is akin to what happens when you run `perl` or `python`, it just sits there waiting for STDIN to complete with an `<EOF>`, then runs the code. It's identical to what `non-interactive` means at the moment.

`tty` is the default and is really only presented to the user for completeness' sake.

`process` is the same as `interactive` except that it explicitly refers to the expectation that STDIN and STDOUT will *not* be TTY. This would resolve the ambiguity that occurs with this patch where `is-interactive` is only set when `interactive` is explictly passed to `repl-mode` regardless of the fact that a regular, default REPL session is also clearly classifiable as interactive.

### DIFF
--- a/src/HLL/Compiler.nqp
+++ b/src/HLL/Compiler.nqp
@@ -13,6 +13,7 @@ class HLL::Compiler does HLL::Backend::Default {
     has %!cli-options;
     has $!backend;
     has $!save_ctx;
+    has $!repl-mode;
 
     method BUILD() {
         # Backend is set to the default one, by default.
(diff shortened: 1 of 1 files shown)


ab5tract:
I've decided to just do the entire set of proposed changes to repl-mode instead of simply proposing said changes :)

patrickbkr (approved):
Let's do this!

PR #830, opened 2024-11-17
[JVM] Fix nqp::attrinited for attribute set to null

usev6:
It was wrong to use nqp::isnull for this check with ba4dc47799. An attribute that has been set to nqp::null is still initialized.


usev6:
Oh, it looks like this change depends on https://github.com/Raku/nqp/pull/829.

Converting to a draft for now.

usev6:
This should be good to merge now.

PR #833, opened 2024-11-29
[JVM] Return exit code for run failure

usev6:
This adjusts the behaviour of the JVM backend to the change from https://github.com/rakudo/rakudo/commit/81c7064ba7. The error callback now accepts a second argument with an exit code.

At least for IOExceptions the original error code from the JVM can be seen in the error message. I don't see a real benefit in extracting that error code and passing it on. Instead, we'll just set the exit code to -1 (which is also expected by at least one spectest: https://github.com/Raku/roast/blob/3032bcf423/S29-os/system.t#L47).


ab5tract (approved):
Looks good to me!

PR #834, opened 2024-12-17
Explicitly deactivate repl-mode for -e

ab5tract:
This fixes some issues that were cropping up in Windows CI runs.


patrickbkr:
Do you have an idea why this only affects Windows?

ab5tract:
> Do you have an idea why this only affects Windows?

Unfortunately not :(

My instincts point towards Windows CI runners being one (or more) degrees less conditioned towards providing a "tty-compatible" substrate, but that's strictly an assumption.

PR #835, opened 2024-12-26
Verify some sizes in qast compiler mast

timo:
If any of the section offsets overflow the 32bit offset number that we write to the file, we throw an exception.

If any frame has more than 65534 locals or lexicals, we throw an exception.
If there are more than 65534 frames, we throw an
exception as well.

Fixes https://github.com/rakudo/rakudo/issues/5733

(unless we want to make that issue about the limitation to locals and lexicals being "what fits into 16bit", and not the too-late and rather-unhelpful failure mode)

### DIFF
--- a/src/vm/moar/QAST/QASTCompilerMAST.nqp
+++ b/src/vm/moar/QAST/QASTCompilerMAST.nqp
@@ -2509,30 +2509,37 @@ class MoarVM::BytecodeWriter {
         $!mbc.write_uint32($offset); # Offset of SC dependencies table
         $!mbc.write_uint32(my uint $num_sc_handles := nqp::elems(@sc_handles)); # Number of entries in SC dependencies table
         $offset := $offset + $sc_deps_size;
+        nqp::die("MAST::CompUnit file size limit reached!") if $offset < $sc_deps_size;
 
         $!mbc.write_uint32($offset); # Offset of extension ops table
         $!mbc.write_uint32($num_extops); # Number of entries in extension ops table
         $offset := $offset + $extops_size;
+        nqp::die("MAST::CompUnit file size limit reached!") if $offset < $extops_size;
 
         $!mbc.write_uint32($offset); # Offset of frames segment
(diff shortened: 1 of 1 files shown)


src/vm/moar/QAST/QASTCompilerMAST.nqp
```diff
@@ -2509,30 +2509,37 @@ class MoarVM::BytecodeWriter {
         $!mbc.write_uint32($offset); # Offset of SC dependencies table
         $!mbc.write_uint32(my uint $num_sc_handles := nqp::elems(@sc_handles)); # Number of entries in SC dependencies table
         $offset := $offset + $sc_deps_size;
+        nqp::die("MAST::CompUnit file size limit reached!") if $offset < $sc_deps_size;
```
MasterDuke17, line 2512: How useful would it be to know which of these checks triggered? E.g., `
        nqp::die("MAST::CompUnit file size limit reached because of number of entries in SC dependencies table!") if $offset < $sc_deps_size;`

src/vm/moar/QAST/QASTCompilerMAST.nqp
```diff
@@ -2761,6 +2775,10 @@ class MoarVM::BytecodeWriter {
         self.align;
         self.write_extops;
         self.align;
+
+        nqp::die("Cannot have more than 65534 MAST::Frame in a CompUnit. This CU has " ~ nqp::elems(@!frames))
```
MasterDuke17, line 2779: Why the differences (65535 vs 65534 and 0xffff vs 65535) from the previous checks?

src/vm/moar/QAST/QASTCompilerMAST.nqp
timo: Not sure it'll be terribly useful to an end user, especially since it may be a small thing that just barely pushes it over the limit.

src/vm/moar/QAST/QASTCompilerMAST.nqp
timo: thanks, i unified the messages

MasterDuke17 (approved):

PR #838, opened 2025-02-17
Use a local for a somewhat hot class member access

MasterDuke17:


lizmat:
For this to be beneficial, it would mean that the majority of cases it would find the key?  Is that the case?

MasterDuke17:
> For this to be beneficial, it would mean that the majority of cases it would find the key? Is that the case?

Seems to be. There were ~220k `getattr`s before and ~115k after.

PR #852, opened 2026-06-06
HLL::Compiler.compile: balance backend per-compile state on early exit

ugexe:
HLL::Backend.start (MoarVM) pushes a frame list onto %COMPILING<moar><frames> at every compile's 'start' stage. The matching pop runs at the 'mast' stage in QASTCompilerMAST.to_mast. A compile that stops before 'mast' (e.g. with `:target('ast')`), or one whose stage call throws, leaves the push dangling. The next nested compile's MAST migration then attaches its frames to the dangling level rather than the outer compile's, leaving those MAST::Frames unreachable from any compunit's @!frames. The outer's deserialization code dies later with `Could not find frame <unit>`.

Snapshot the backend's per-compile state before the stages loop and rebalance it afterwards via two optional backend hooks (compile_snapshot, compile_cleanup). Wrap each stage call in CATCH so the cleanup also runs when a stage throws, capture the exception, exit the stages loop, then rethrow with its original payload.

The legacy Rakudo frontend never stops short of 'mast' and never throws across the stages loop boundary, so on every existing non-RakuAST compile the snapshot matches the post-loop depth, the cleanup pop loop runs no iterations, and the caught variable stays null.

### DIFF
--- a/src/HLL/Compiler.nqp
+++ b/src/HLL/Compiler.nqp
@@ -503,6 +503,17 @@ class HLL::Compiler does HLL::Backend::Default {
             nqp::unshift(@stages, 'start');
         }
 
+        # Backends may keep per-compile state on %*COMPILING (e.g. a frame
+        # list that the 'start' stage pushes for the 'mast' stage to pop).
+        # Snapshot that state so compile_cleanup can rebalance it if the
+        # stages loop exits before the consuming stage runs, whether via a
+        # 'target' adverb shorter than the final stage or an in-flight
+        # exception.
+        my $compile_snapshot := nqp::can($!backend, 'compile_snapshot')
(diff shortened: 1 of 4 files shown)


src/HLL/Compiler.nqp
```diff
@@ -532,6 +556,11 @@ class HLL::Compiler does HLL::Backend::Default {
             last if $_ eq $target;
         }
 
+        nqp::can($!backend, 'compile_cleanup')
+            && $!backend.compile_cleanup($compile_snapshot);
+
+        nqp::rethrow($caught) if !nqp::isnull($caught);
```
lizmat, line 562: `unless nqp::isnull($caught)` ?

PR #854, opened 2026-07-14
Release dispatch op arg registers under their coerced kind

ugexe:
Previously add-dispatcher-op recorded each argument's register kind before coercing small-int, uint, and num32 arguments to full-width registers, then released the coerced register under that stale kind. That put an int64 register on the uint64 free-list (and similar for the other coerced kinds), so a later fresh_register of that kind in the same frame handed out a register of the wrong type and produced invalid bytecode, e.g. "operand type does not match register type for op decont_u" compiling a Raku sub that passes a uint parameter to nqp::dispatch.

Release with the register's post-coercion kind instead, the same way the call op's argument handling already does.

### DIFF
--- a/src/vm/moar/QAST/QASTOperationsMAST.nqp
+++ b/src/vm/moar/QAST/QASTOperationsMAST.nqp
@@ -1888,7 +1888,6 @@ my sub add-dispatcher-op($qastcomp, $op, :$prefix) {
     # Compile arguments and form callsite.
     my @arg_mast;
     my @arg_idxs;
-    my @arg_kinds := nqp::list_i;
     my $regalloc := $qastcomp.regalloc;
     my $frame := $qastcomp.mast_frame;
     for @args -> $arg {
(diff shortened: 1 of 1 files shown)


ugexe:
```
# Legacy frontend
➜  nqp git:(ugexe/dispatch-op-arg-release-kind) raku -e 'use nqp; sub f(uint $x) { nqp::dispatch("boot-value", $x) }; say f(3)'

3


# RakuAST frontend (before this PR)
➜  nqp git:(ugexe/dispatch-op-arg-release-kind) RAKUDO_RAKUAST=1 raku -e 'use nqp; sub f(uint $x) { nqp::dispatch("boot-value", $x) }; say f(3)'

Bytecode validation error at offset 50, instruction 8:
operand type 160 (?) does not match register type 32 (int64) for op decont_u in frame f
  in block <unit> at -e line 1


# RakuAST frontend (after this PR)
➜  nqp git:(ugexe/dispatch-op-arg-release-kind) RAKUDO_RAKUAST=1 ../install/bin/raku -e 'use nqp; sub f(uint $x) { nqp::dispatch("boot-value", $x) }; say f(3)'

3
```

ugexe:
The bad release happens on both frontends. RakuAST is the one that hits it because it compiles the body first and the param prologue after. The prologue for a native uint parameter asks for a fresh uint register and gets handed the mislabeled int register. Legacy never asks for a uint register after the dispatch in that frame, so it works by accident. The call op already releases args with their post coercion kind. This makes dispatch do the same.

PR #856, opened 2026-08-05
Fix lookbehind for patterns that cannot be flipped

### LINKED ISSUE #6087: `after` incorrectly handles (flips) "sub-rules"
```
my regex ab { ab }
say .pos given 'ab ba' ~~ / <?after    ab   > /; # 2
say .pos given 'ab ba' ~~ / <?after   <ab>  > /; # 5 <- WRONG!
say .pos given 'ab ba' ~~ / <?after <{'ab'}>> /; # 5 <- WRONG!
say .pos given 'ab ba' ~~ / <?after <{'ba'}>> /; # 2 <- WRONG!
```

[A 2012 Rakudo commit "Implement lookbehind" ](https://github.com/Raku/nqp/commit/3442c181da4df6b57ffa8e90d8f049ddeafe2d7b) involves conditional flipping of qast. That's got to be related inasmuch as something related to that flipping has gone wrong, either from the getgo (though jnthn wrote it, so I doubt it's that), or at some point since.

(I arrived at this golf, and my hint about what seems to be the problem, as an evolution of my initial answer to SO [Why doesn't this look-behind assertion with a predefined regular expression work as expected?](https://stackoverflow.com/a/79906200/1077672).)

ugexe:
The <?after ...> assertion compiles by reversing the pattern syntax tree and matching it against a reversed copy of the target string. flip_ast only knows how to reverse some nodes. Subrule calls (named rules, interpolated patterns, backreferences), embedded code, and zero-width checks other than anchors were left matching left-to-right against the reversed target, giving wrong results:

    my regex ab { ab }
    say .pos given 'ab ba' ~~ / <?after <ab> > /;  # 5, should be 2

Previously the reversal was unconditional. Now the actions only reverse when can_flip_ast confirms every node in the pattern is one flip_ast handles. Otherwise the call is emitted against the new after_scan method, which matches the unmodified pattern forward from each position at or before the current one and succeeds if any match ends exactly at the current position. A match that ends elsewhere is backtracked into so a pattern like 'a || ab' can still find the ending that lines up.

The after method now also passes or fails on a fresh cursor rather than the one the regex ran on. That cursor held the reversed target and a $!from in reversed-string coordinates, which made the capturing <after ...> form produce a negative-width match object.

P5Regex lookbehinds get the same treatment. Its flip_ast only reverses literals and concat order, so its can_flip_ast additionally sends anchors down the scanning path.

Helps resolve rakudo/rakudo#6087

### DIFF
--- a/src/NQP/Actions.nqp
+++ b/src/NQP/Actions.nqp
@@ -1925,8 +1925,12 @@ class NQP::RegexActions is QRegex::P6Regex::Actions {
                         $qast[0].push(self.qbuildsub($<nibbler>.ast, :anon(1), :addself(1)));
                         $qast[0].push(QAST::IVal.new( :value($litlen) ));  # optional offset to before
                     }
-                    else {
+                    elsif self.can_flip_ast($<nibbler>.ast) {
                         $qast[0].push(self.qbuildsub(self.flip_ast($<nibbler>.ast), :anon(1), :addself(1)));
+                    }
+                    else {
+                        $qast[0][0].value('after_scan');
+                        $qast[0].push(self.qbuildsub($<nibbler>.ast, :anon(1), :addself(1)));
                     }
                 }
                 else {
--- a/src/QRegex/Cursor.nqp
+++ b/src/QRegex/Cursor.nqp
@@ -1132,16 +1132,70 @@ role NQPMatchRole is export {
         ) if nqp::isnull_s($flipped);
 
         # Set up cursor with  updated shared info with flipped string
-        my $cursor := self."!cursor_start_cur"();
+        my $flipped_cursor := self."!cursor_start_cur"();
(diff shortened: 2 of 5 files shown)


coke:
Failing tests on windows.

lizmat:
> Failing tests on windows.

@coke isn't that a more general issue atm?

coke:
Ah, timo fixed the failures on MoarVM, I didn't verify nqp at the time. Last time I checked, we were only failing JVM tests. (a few months ago?).

So, not needed for this one particularly, but we do need to fix that.

coke:
Can you rebase this (and the other one from 3 weeks ago) against main so the windows CI can be rechecked?

coke:
🎉

PR #858, opened 2026-08-05
Update windows target for CI

coke:


lizmat:
@coke: Looking forward to the results!

coke:
Looks back to the previous expected JVM failures, the 2 win MVM jobs are passing.

PR #859, opened 2026-08-05
Add %a/%A (C99 hex float) directives to HLL sprintf

dankogai:
C99 Floating-point notation support has been long overdue. Perl 5 has done so in 5.22 in 2015, more than 11 years ago.

This adds C99 hexadecimal floating-point notation to the HLL sprintf:

```
nqp::sprintf('%a',    [0.1])   # 0x1.999999999999ap-4
nqp::sprintf('%A',    [0.1])   # 0X1.999999999999AP-4
nqp::sprintf('%.3a',  [3.14159265358979])  # 0x1.922p+1
nqp::sprintf('%020a', [1.5])   # 0x0000000000001.8p+0
```

Supports the usual flags (`+`, space, `-`, `0`, `#`), width and precision, including `*` for both.

## Implementation

Pure NQP and backend-portable — no IEEE bit-twiddling, only bigint ops already used elsewhere in sprintf.nqp. The value is scaled into [1, 2) by exact power-of-two operations so that mantissa × 2**52 is an exact 53-bit integer; its base-16 digits are the leading digit plus exactly 13 hex fraction digits. Precision rounding is done in bigint arithmetic, correctly rounded with IEEE ties-to-even.

Semantics, verified against C implementations:
- exact rounding ties go to even, matching glibc (macOS gdtoa truncates ties instead; being pure NQP, output here is identical on all platforms)
- subnormals print normalized (`5e-324` → `0x1p-1074`), like BSD and Perl, rather than glibc's `0x0.…` denormalized form
- zero padding goes between the `0x` prefix and the mantissa; `#` preserves the radix point (both as in C)
- `Inf`/`-Inf`/`NaN`, consistent with `%e`/`%f`/`%g`

## Tests

37 new tests in t/hll/06-sprintf.t (basics, precision/rounding/carry, ties, flags, padding, ±0, subnormals, max double, Inf/NaN). The test that asserted `%a` is an *invalid* directive now uses `%v`. Full `make test` passes (13513 tests).

## Follow-up

A rakudo PR adding C99 hexfloat *literals* (`0x1.8p+1`) to both frontends is ready to follow once this lands; with both in place, hexfloat literals round-trip through `sprintf('%a', …)`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/HLL/sprintf.nqp
+++ b/src/HLL/sprintf.nqp
@@ -24,6 +24,7 @@ my module sprintf {
         }
 
         proto token directive { <...> }
+        token directive:sym<a> { '%' <idx>? <flags>* <size>? [ '.' <precision=.size> ]? $<sym>=<[aA]> }
         token directive:sym<b> { '%' <idx>? <flags>* <size>? [ '.' <precision=.size> ]? $<sym>=<[bB]> }
         token directive:sym<c> { '%' <idx>? <flags>* <size>? <sym> }
         token directive:sym<d> { '%' <idx>? <flags>* <size>? [ '.' <precision=.size> ]? $<sym>=<[di]> }
@@ -150,7 +151,8 @@ my module sprintf {
         sub padding_char($st) {
             my $padding_char := ' ';
             if !$st<precision> && !is_minus($st)
-            || $st<sym> ~~ /<[eEfFgG]>/ {
+            || $st<sym> ~~ /<[eEfFgG]>/
+            || !is_minus($st) && $st<sym> ~~ /<[aA]>/ {
                 $padding_char := '0' if $_<zero> for $st<flags>;
             }
             $padding_char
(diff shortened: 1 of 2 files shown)


lizmat:
@dankogai when you're working on the Rakudo version, please also look at the src/core.e/Formatter.rakumod logic!

dankogai:
@lizmat thanks and long time no see. Done — `%a`/`%A` are now implemented in `src/core.e/Formatter.rakumod` (+ `Formatter/Grammar.rakumod`) as well, using the same algorithm: mantissa digits as an exact integer, precision rounding in integer arithmetic with ties-to-even, zero-padding between the `0x` prefix and the mantissa. Output is byte-identical with this nqp implementation for all finite values; NaN/Inf follow Formatter's existing conventions for the float directives (as `%e`/`%f`/`%g` already do). There are 31 tests running under `use v6.e.PREVIEW`.

The work is staged on my fork together with the hexfloat *literal* support for both frontends: dankogai/rakudo#1. Once this PR lands I'll bump `NQP_REVISION` and submit that as the follow-up rakudo PR.

dankogai:
BTW JVM CIs are failing not because of this PR but because of the memory shortage of the testbeds :(
https://dev.azure.com/Raku/nqp/_build/results?buildId=68&view=logs&jobId=ab856b8b-876f-5746-08e5-94ac68404358&j=ab856b8b-876f-5746-08e5-94ac68404358&t=26b600de-420e-5a1b-fdba-a0a236382af3

lizmat:
@dankogai: yes, this is a known issue.  Combined with the fact that the JVM backend hasn't had any  TLC for year now, so it has quite bitrotted  :-(

lizmat:
> The work is staged on my fork together with the hexfloat literal support for both frontends: https://github.com/dankogai/rakudo/pull/1. Once this PR lands I'll bump NQP_REVISION and submit that as the follow-up rakudo PR.

@dankogai: The NQP bump has already been done, so you don't have to worry about that.

